PR #43, opened 2026-04-14
feat(a2a-client): make A2AClient generic over Transport to enable zero-cost static dispatch

### LINKED ISSUE #42: Make A2AClient generic over Transport to enable zero-cost static dispatch
## Problem

`A2AClient` currently stores its transport as `Box<dyn Transport>`, which forces a heap allocation and vtable indirection on every call — even when the concrete transport type is known at compile time.

```rust
pub struct A2AClient {
    transport: Box<dyn Transport>,
    // ...
}

impl A2AClient {
    pub fn new(transport: Box<dyn Transport>) -> Self { ... }
}
```

This means callers who know their transport type statically
(e.g. JsonRpcTransport, RestTransport) still pay for dynamic dispatch unnecessarily.

## Proposed solution

Make A2AClient generic over T: Transport:

```rust
pub struct A2AClient<T: Transport> {
    transport: T,
    // ...
}

impl<T: Transport> A2AClient<T> {
    pub fn new(transport: T) -> Self { ... }
}
```

This enables two usage paths:

- Static dispatch (zero-cost): When the transport type is known at compile time, the compiler monomorphizes and can inline through transport calls.

```rust
let client = A2AClient::new(JsonRpcTransport::new(http_client, url));
// type: A2AClient<JsonRpcTransport> — no boxing, no vtable
```  

- Dynamic dispatch (runtime negotiation): When the transport is selected at runtime via A2AClientFactory::create_from_card(), the return type becomes A2AClient<Box<dyn Transport>>, preserving the existing factory-based protocol negotiation.

```rust
let client = factory.create_from_card(&card).await?;
// type: A2AClient<Box<dyn Transport>> — dynamic dispatch as before
```

msardara:
Fixes #42

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -21,6 +21,7 @@ dependencies = [
  "agntcy-a2a",
  "agntcy-a2a-pb",
  "async-trait",
+ "auto_impl",
  "bytes",
  "futures",
  "pin-project-lite",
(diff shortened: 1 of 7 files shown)


muscariello (approved):
LGTM

PR #44, opened 2026-04-14
ci: simplify CI workflow setup

msardara:
### Summary

- Replace `dtolnay/rust-toolchain` + `Swatinem/rust-cache` with
  `actions-rust-lang/setup-rust-toolchain` (v1.16.0), which bundles
  both toolchain installation and build caching in a single step
- Remove `arduino/setup-protoc` step from all workflows — protoc
   is already vendored at build time via the `protoc-bin-vendored`
   workspace dependency
- Remove explicit `RUSTFLAGS: "-D warnings"` from `ci.yml` env block
   as `setup-rust-toolchain` sets this by default

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,6 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
-  RUSTFLAGS: "-D warnings"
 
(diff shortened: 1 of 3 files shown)


muscariello (approved):
LGTM

PR #46, opened 2026-04-15
feat: add stdio transport binding

### LINKED ISSUE #15: task: add stdio transport binding
Type: Task

## Summary
Add a stdio transport binding for subprocess-hosted agents using stdin/stdout, alongside the current JSON-RPC over HTTP, REST / HTTP+JSON, gRPC, and SSE streaming support.

## Why
The existing bindings are network-oriented. A stdio transport would make local subprocess execution a first-class option for tool-style integrations, local development, and process-scoped agent hosting while still reusing the core transport abstraction and shared protocol types.

## Scope / plan
- [ ] Add a stdio protocol identifier and interface contract for subprocess-backed agents
- [ ] Implement a client transport that spawns and manages a subprocess over stdin/stdout
- [ ] Define request framing, startup handshake, streaming behavior, shutdown, and cancellation semantics
- [ ] Define how ServiceParams are carried without HTTP or gRPC metadata
- [ ] Define A2AError mapping for process exits, broken pipes, invalid frames, and timeouts
- [ ] Add integration tests and supporting docs for the stdio contract and lifecycle behavior

## Acceptance criteria
- A client can launch a subprocess-backed agent and complete unary and streaming operations over stdio
- Lifecycle handling is deterministic for startup failure, clean exit, cancellation, and broken-pipe cases
- ServiceParams propagation and A2AError mapping are specified and covered by tests
- Docs describe the stdio contract clearly enough for another implementation to interoperate

Parent feature: #12

Ketankhunti:
## Summary

Adds a new `a2a-stdio` crate that implements the A2A STDIO transport binding - a subprocess-based transport for local agent communication using LSP-style Content-Length framing over stdin/stdout.

Fixes: #15

## Motivation

The A2A spec defines multiple transport bindings. The existing crate supports HTTP/REST, JSON-RPC, gRPC, and SLIMRPC. This PR adds STDIO as a Custom Protocol Binding (CPB) for local, single-connection scenarios where spawning a subprocess is simpler than standing up a network server.

## Design

Based on the design agreed in #15 (Option A):

- **Wire format:** LSP-style `Content-Length: N\r\n\r\n<body>` framing
- **Protocol:** JSON-RPC 2.0 with slash-separated method names (`message/send`, `tasks/get`, etc.)
- **Handshake:** Server sends `handshake` → client replies with `handshakeAck` (accept/reject + variant selection)
- **Session ID:** Via `A2A_SESSION_ID` environment variable or auto-generated UUIDv7
- **Streaming:** Events sent as JSON-RPC notifications (no `id`), followed by a final response with the original request `id`

### New crate: `a2a-stdio`

| Module | Description |
|--------|-------------|
| `errors.rs` | `StdioError` enum: `Io`, `Json`, `InvalidHeader`, `HandshakeFailed`, `Closed` |
| `framing.rs` | `read_frame()` / `write_frame()` for Content-Length framing |
| `handshake.rs` | `Handshake`, `HandshakeAck`, `HandshakeFeatures` types + read/write I/O |
| `client.rs` | `StdioTransport` (spawns subprocess, implements `Transport` trait), `StdioTransportFactory` |
| `server.rs` | `StdioServer<H: RequestHandler>` with `dispatch_unary!` / `dispatch_streaming!` macros, `serve()` convenience function |

### Other changes

- `a2a/src/types.rs`: Added `TRANSPORT_PROTOCOL_STDIO` constant
- `Cargo.toml` (workspace root): Added `a2a-stdio` to members and workspace dependencies
- `Cargo.lock`: Updated with new crate entry

## Testing

- **19 unit tests**: framing (6), handshake (9), client URL parsing (4)
- **14 e2e integration tests**: Uses `tokio::io::duplex` to wire server ↔ client in-process without spawning a subprocess. Covers all 11 `RequestHandler` methods, streaming, error paths, handshake reject, EOF handling, and sequential requests.
- Full `cargo test --workspace` passes with zero failures
- `cargo clippy --workspace --all-targets` clean

### DIFF
--- a/a2a-stdio/Cargo.toml
+++ b/a2a-stdio/Cargo.toml
@@ -11,6 +11,12 @@ rust-version.workspace = true
 [lib]
 name = "a2a_stdio"
 
+# Test-only helper subprocess used by integration tests in tests/spawned.rs.
+# Not part of the public surface; kept under tests/bin/ to make that explicit.
+[[bin]]
+name = "stdio_test_helper"
+path = "tests/bin/stdio_test_helper.rs"
+
 [dependencies]
 a2a = { workspace = true }
 a2a-client = { workspace = true }
--- a/a2a-stdio/Cargo.toml
+++ b/a2a-stdio/Cargo.toml
@@ -16,9 +16,11 @@ a2a = { workspace = true }
 a2a-client = { workspace = true }
 a2a-server = { workspace = true }
 tokio = { workspace = true }
+tokio-stream = { workspace = true }
 futures = { workspace = true }
 async-trait = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
 thiserror = { workspace = true }
 tracing = { workspace = true }
+uuid = { workspace = true }
--- a/a2a-stdio/Cargo.toml
+++ b/a2a-stdio/Cargo.toml
@@ -1,0 +1,24 @@
+[package]
+name = "agntcy-a2a-stdio"
+version = "0.1.0"
+description = "A2A v1 STDIO transport binding for client and server"
+readme = "README.md"
+edition.workspace = true
+license.workspace = true
+repository.workspace = true
+rust-version.workspace = true
+
+[lib]
+name = "a2a_stdio"
+
+[dependencies]
+a2a = { workspace = true }
+a2a-client = { workspace = true }
+a2a-server = { workspace = true }
+tokio = { workspace = true }
+futures = { workspace = true }
+async-trait = { workspace = true }
+serde = { workspace = true }
+serde_json = { workspace = true }
+thiserror = { workspace = true }
+tracing = { workspace = true }
--- a/a2a-stdio/src/client.rs
+++ b/a2a-stdio/src/client.rs
@@ -18,7 +18,7 @@ use tokio_stream::wrappers::ReceiverStream;
 
 use crate::errors::StdioError;
 use crate::framing;
-use crate::handshake::{self, Handshake, HandshakeAck};
+use crate::handshake::{self, HandshakeAck};
 
 // ---------------------------------------------------------------------------
 // Stdio method names (slash-separated, per maintainer design doc)
@@ -53,8 +53,6 @@ pub struct StdioTransport {
     writer: Arc<Mutex<Box<dyn AsyncWrite + Send + Unpin>>>,
     /// Reader from the child's stdout (framed JSON-RPC).
     reader: Arc<Mutex<BufReader<tokio::process::ChildStdout>>>,
-    /// The handshake received from the server.
-    _handshake: Handshake,
 }
 
 impl StdioTransport {
@@ -100,11 +98,14 @@ impl StdioTransport {
         let ack = HandshakeAck::accept(selected);
         handshake::write_handshake_ack(&mut writer, &ack).await?;
 
+        // Handshake content (selected variant, features) is currently unused
+        // beyond the accept/reject decision; drop it.
+        let _ = hs;
+
         Ok(StdioTransport {
             child: Arc::new(Mutex::new(child)),
             writer: Arc::new(Mutex::new(writer)),
             reader: Arc::new(Mutex::new(reader)),
-            _handshake: hs,
         })
     }
 
@@ -440,6 +441,13 @@ impl TransportFactory for StdioTransportFactory {
 /// - `stdio:///path/to/binary` → program="/path/to/binary", args=[]
 /// - `stdio:///path/to/binary?arg1&arg2` → program="/path/to/binary", args=["arg1", "arg2"]
 /// - `/path/to/binary arg1 arg2` → program="/path/to/binary", args=["arg1", "arg2"]
+///
+/// # Limitations
+///
+/// Arguments in the `stdio://` form are split on `&` and the path is split on
+/// the first `?`. As a result, individual arguments cannot themselves contain
+/// the `&` or `?` characters. For commands that need such arguments, use the
+/// plain `program arg1 arg2` form instead.
 fn parse_stdio_url(url: &str) -> Result<(String, Vec<String>), A2AError> {
     if let Some(rest) = url.strip_prefix("stdio://") {
         let (path, query) = rest.split_once('?').unwrap_or((rest, ""));
--- a/a2a-stdio/src/client.rs
+++ b/a2a-stdio/src/client.rs
@@ -189,9 +189,12 @@ impl StdioTransport {
             return Err(A2AError::new(err.code, err.message));
         }
 
-        let result = rpc_response
-            .result
-            .ok_or_else(|| A2AError::internal("response missing result"))?;
+        // For methods returning `()` the JSON-RPC response carries
+        // `"result": null`, which serde deserializes into `None` for
+        // `Option<Value>`. Treat a missing/null result as `Value::Null`
+        // so unit responses round-trip correctly while non-null results
+        // still flow through normal deserialization.
+        let result = rpc_response.result.unwrap_or(serde_json::Value::Null);
 
         serde_json::from_value(result)
             .map_err(|e| A2AError::internal(format!("deserialize result: {e}")))
--- a/a2a-stdio/src/client.rs
+++ b/a2a-stdio/src/client.rs
@@ -11,6 +11,7 @@ use futures::stream::BoxStream;
 use serde::Serialize;
 use serde::de::DeserializeOwned;
 use std::sync::Arc;
+use std::time::Duration;
 use tokio::io::{AsyncWrite, AsyncWriteExt, BufReader};
 use tokio::process::{Child, Command};
 use tokio::sync::Mutex;
@@ -19,6 +20,13 @@ use tokio_stream::wrappers::ReceiverStream;
 use crate::errors::StdioError;
 use crate::framing;
 use crate::handshake::{self, HandshakeAck};
+
+/// How long to wait for the subprocess to send its initial handshake.
+const HANDSHAKE_TIMEOUT: Duration = Duration::from_secs(10);
+
+/// How long to wait for the subprocess to exit after closing its stdin
+/// before forcibly killing it.
+const CLOSE_TIMEOUT: Duration = Duration::from_secs(5);
 
 // ---------------------------------------------------------------------------
 // Stdio method names (slash-separated, per maintainer design doc)
@@ -46,6 +54,16 @@ mod stdio_methods {
 ///
 /// The client spawns the child process, performs the handshake, then sends
 /// JSON-RPC requests over the child's stdin and reads responses from stdout.
+/// A transport that communicates with an agent subprocess over stdio pipes.
+///
+/// The client spawns the child process, performs the handshake, then sends
+/// JSON-RPC requests over the child's stdin and reads responses from stdout.
+///
+/// # Concurrency
+///
+/// Each call to [`Transport`] takes a request-level lock that is held across
+/// the full write/read cycle. Concurrent calls therefore block rather than
+/// race on the underlying pipes — responses are not multiplexed by `id`.
 pub struct StdioTransport {
     /// The child process. Held for lifetime management.
     child: Arc<Mutex<Child>>,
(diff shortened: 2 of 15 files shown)


a2a-stdio/src/server.rs
```diff
@@ -0,0 +1,254 @@
...
+
+        // Step 3: Main request loop.
+        let handler = &*self.handler;
+        loop {
+            let frame = match framing::read_frame(&mut reader).await? {
+                Some(f) => f,
+                None => break, // EOF — client closed stdin.
+            };
+
+            let value: serde_json::Value = serde_json::from_slice(&frame)?;
```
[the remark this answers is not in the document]
Ketankhunti: Addressed in 2ab6d45 - replaced ? with a match that sends a PARSE_ERROR response and continues the loop.

a2a-stdio/src/client.rs
```diff
@@ -0,0 +1,481 @@
...
+                            break;
+                        }
+                    };
+
+                    if let Some(err) = rpc_response.error {
+                        let _ = tx.send(Err(A2AError::new(err.code, err.message))).await;
+                    } else if let Some(result) = rpc_response.result {
+                        if let Ok(sr) = serde_json::from_value::<StreamResponse>(result) {
+                            let _ = tx.send(Ok(sr)).await;
+                        }
```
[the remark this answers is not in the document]
Ketankhunti: Addressed in 2ab6d45 - non-null malformed results now explicitly report an error through the channel instead of being silently dropped.

msardara (commented):
**Nit (non-blocking):** `StdioTransport` doesn't document that concurrent requests are unsafe. Since the server processes requests sequentially and the client doesn't match response `id`s back to requests, concurrent calls to `call()` / `call_streaming()` would race on the reader.

Consider adding a doc comment on the `StdioTransport` struct (`client.rs:51`):

```rust
/// **Note:** This transport does not support concurrent requests. The server
/// processes requests sequentially, so callers must await each request before
/// sending the next. Concurrent calls will race on the reader and may receive
/// each other's responses.
```

msardara (commented):
Great job @Ketankhunti! I reviewed the PR and found few issues, but for the rest the PR looks OK.

The major one is that `StdioTransport` doesn't guard against concurrent requests. Since the server processes requests sequentially and the client doesn't match response `id`s back to requests, concurrent calls to `call()` / `call_streaming()` would race on the reader. This might not be an issue as we have only one client by design, but the API does not enforce issuing 1 request at a time, as nothing stops the caller from calling send_message and get_task concurrently.

A few options to address this:

1. **Document the constraint (weak)** — add a doc comment on `StdioTransport` warning that callers must serialize requests.
2. **Add a request-level lock** — wrap the full write+read cycle in a single `Mutex` so concurrent calls block rather than race. Cheap and prevents misuse.
3. **Support true multiplexing** — add a background reader task that routes responses by `id` to pending request futures. More complex but enables concurrent requests.

a2a-stdio/src/framing.rs
```diff
@@ -0,0 +1,149 @@
...
+
+        if let Some(value) = trimmed.strip_prefix(CONTENT_LENGTH_PREFIX) {
+            content_length = Some(value.parse::<usize>().map_err(|_| {
+                StdioError::InvalidHeader(format!("invalid Content-Length value: {value}"))
+            })?);
+        }
+        // Ignore unknown headers (e.g. Content-Type) for forward compatibility.
+    }
+
+    let length = content_length
```
msardara, line 62: length comes directly from the peer's Content-Length header with no upper bound. A buggy or malicious subprocess could send Content-Length: 9999999999 and OOM the process.

Consider adding a max frame size check here:

```rust
const MAX_FRAME_SIZE: usize = 64 * 1024 * 1024; // 64 MB
if length > MAX_FRAME_SIZE {
    return Err(StdioError::InvalidHeader(format!("Content-Length {length} exceeds max {MAX_FRAME_SIZE}")));
}
```

a2a-stdio/Cargo.toml
```diff
@@ -0,0 +1,26 @@
+[package]
+name = "agntcy-a2a-stdio"
+version = "0.1.0"
+description = "A2A v1 STDIO transport binding for client and server"
+readme = "README.md"
```
msardara, line 5: This references a README.md that isn't included in the PR. Either add the file or remove this line to avoid cargo publish warnings.

a2a-stdio/src/client.rs
```diff
@@ -0,0 +1,495 @@
...
+
+// ---------------------------------------------------------------------------
+// StdioTransport
+// ---------------------------------------------------------------------------
+
+/// A transport that communicates with an agent subprocess over stdio pipes.
+///
+/// The client spawns the child process, performs the handshake, then sends
+/// JSON-RPC requests over the child's stdin and reads responses from stdout.
+pub struct StdioTransport {
```
msardara, line 67: StdioTransport doesn't guard against concurrent requests. The server processes requests sequentially and the client doesn't match response ids back to requests, so concurrent calls to call() / call_streaming() would race on the reader. Options:
1. Document the constraint with a doc comment warning callers must serialize requests.
2. Wrap the full write+read cycle in a single Mutex so concurrent calls block rather than race.
3. Support true multiplexing with a background reader task routing responses by id.

a2a-stdio/src/client.rs
```diff
@@ -0,0 +1,495 @@
...
+        let stdout = child
+            .stdout
+            .take()
+            .ok_or_else(|| StdioError::HandshakeFailed("failed to open child stdout".into()))?;
+
+        let mut reader = BufReader::new(stdout);
+        let mut writer: Box<dyn AsyncWrite + Send + Unpin> = Box::new(stdin);
+
+        // Server sends handshake first.
+        let hs = handshake::read_handshake(&mut reader).await?;
```
msardara: Should we set a timeout for this handshake? If the subprocess hangs, this will block forever.

a2a-stdio/src/client.rs
```diff
@@ -0,0 +1,495 @@
...
+            writer
+                .shutdown()
+                .await
+                .map_err(|e| A2AError::internal(format!("shutdown stdin: {e}")))?;
+        }
+
+        // Wait for child to exit.
+        let mut child = self.child.lock().await;
+        child
+            .wait()
```
msardara: This may block forever as well if the child does not exit. Please consider adding a timeout here as well and kill() the child as fallback.

a2a-stdio/src/client.rs
```diff
@@ -0,0 +1,495 @@
...
+/// JSON-RPC requests over the child's stdin and reads responses from stdout.
+pub struct StdioTransport {
+    /// The child process. Held for lifetime management.
+    child: Arc<Mutex<Child>>,
+    /// Writer to the child's stdin (framed JSON-RPC).
+    writer: Arc<Mutex<Box<dyn AsyncWrite + Send + Unpin>>>,
+    /// Reader from the child's stdout (framed JSON-RPC).
+    reader: Arc<Mutex<BufReader<tokio::process::ChildStdout>>>,
+    /// The handshake received from the server.
+    _handshake: Handshake,
```
msardara: Nit: This field is stored but never read. If it's for future use, a // TODO comment would help. Otherwise remove it.

a2a-stdio/src/client.rs
```diff
@@ -0,0 +1,495 @@
...
+
+        Ok(Box::new(transport))
+    }
+}
+
+/// Parse a stdio URL into program and args.
+///
+/// Supports formats:
+/// - `stdio:///path/to/binary` → program="/path/to/binary", args=[]
+/// - `stdio:///path/to/binary?arg1&arg2` → program="/path/to/binary", args=["arg1", "arg2"]
```
msardara, line 504: Nit: Using & as argument separator means arguments containing & or ? can't be represented. Worth a doc comment noting this limitation.

msardara (changes requested):

muscariello:
@Ketankhunti code coverage needs to be improved from current 64% to 95%

a2a-stdio/src/framing.rs
Ketankhunti: ### `framing.rs`: unbounded frame allocation

> Added `MAX_FRAME_SIZE = 64 MiB` in `20e7cf1`. The length prefix is now validated **before** the `vec![0u8; length]` allocation, so a malicious or corrupt peer can no longer trigger an OOM. Returns `FramingError::FrameTooLarge` instead.

a2a-stdio/src/client.rs
Ketankhunti: ### `StdioTransport`: concurrent request safety

> Fixed in `af03f94`. `StdioTransport` now holds an `Arc<Mutex<()>>` that is acquired for the full write+read cycle of every `call` / `call_streaming`. For streaming, the guard is moved into the background reader task (`lock_owned`) so it's released only after the stream terminates. This serializes request/response framing on the single stdio pipe and prevents interleaving when callers invoke the transport concurrently.

a2a-stdio/src/client.rs
Ketankhunti: ### `client.rs`: unused `_handshake` field

> Removed in `20e7cf1`. The negotiated handshake info wasn't read after construction, so the field is gone.

a2a-stdio/src/client.rs
Ketankhunti: ### `parse_stdio_url`: `&` / `?` limitation

> Documented in `20e7cf1`. Added a doc note that the current parser splits on the first `?` and on `&`, so program paths or args containing those characters need to be passed via `StdioTransportConfig` directly rather than via the URL form. Happy to switch to full percent-decoding in a follow-up if you'd prefer.

a2a-stdio/src/client.rs
Ketankhunti: ### Handshake can hang forever

> Fixed in `af03f94`. The handshake is now wrapped in `tokio::time::timeout(HANDSHAKE_TIMEOUT, ...)` (10s). On timeout we `child.start_kill()` + `child.wait()` so the subprocess is reaped, and return a `TransportError::Other("handshake timeout")`. Covered by new tests in `tests/spawned.rs` (`hang-no-handshake` and `handshake-then-hang` helper modes).

a2a-stdio/src/client.rs
Ketankhunti: ### `child.wait()` in `destroy()` can hang

> Fixed in `af03f94`. `destroy()` first issues a graceful shutdown, then waits with `tokio::time::timeout(CLOSE_TIMEOUT, child.wait())` (5s). On timeout we `start_kill()` and `wait()` again so the child is reaped instead of leaking. Covered by `tests/spawned.rs::destroy_kills_unresponsive_child`.

a2a-stdio/Cargo.toml
Ketankhunti: ### Missing README.md

> Added README.md in `20e7cf1` describing the transport, framing, handshake, and usage with `StdioTransportFactory` / `parse_stdio_url`.

Ketankhunti:
@muscariello  Coverage is now at **96.07%** on the patch (up from 64% on the original PR).

 Pushed three follow-up commits:
> - `0bd45f3` — fix `clippy::useless_conversion` flagged by Rust 1.95 on macOS/Windows.
> - `877e7fc` — apply `rustfmt` (Ubuntu fmt-check was failing).
> - `b3ca336` — additional coverage: the `full-echo` test helper now drives the server through the public `serve()` convenience function, and three new helper modes (`stream-bad-frame`, `stream-bad-notification`, `stream-bad-final`) cover the three error branches in the streaming background reader (`parse frame`, `parse streaming event`, `failed to parse final stream response`).

 Test suite is now 26 unit + 21 e2e + 19 spawned = **66 tests**, all passing; clippy and fmt clean across the workspace.

 The remaining uncovered lines (21 in client.rs, 6 in server.rs) are error arms of `map_err` wrappers around `tokio` I/O calls (`writer.shutdown()`, `child.wait()`, framing read/write) — triggering them requires injecting OS-level pipe failures, which isn't feasible in portable tests. Happy to add mock-IO tests for those if you'd prefer, but I'd argue the cost/benefit isn't favorable for what are essentially `Err(e) => Err(wrap(e))` arms.

Ketankhunti:
Hi @msardara , Did you get a chance to review the PR?

msardara:
Hey @Ketankhunti, not yet but I'll have a look soon ✅ Sorry for the delay.

muscariello:
I need to check this when I am back. No hurry. Like I said we need a custom protocol binding proposal.
There is a whole process to follow to get a new CPB in the project.

muscariello:
@Ketankhunti can you create an issue with a spec proposal following the line of the websocket CPB proposal?
https://github.com/a2aproject/A2A/issues/1864
experimental-cpb-stdio

Ketankhunti:
Hi @muscariello, I will do it, let me first look at the issue.

PR #50, opened 2026-04-23
fix(a2a-client): drain parsed SSE events before re-polling byte stream

### LINKED ISSUE #52: bug: SSE events silently dropped when multiple events arrive in single byte chunk
## Problem

`parse_sse_stream` and `parse_sse_stream_rest` in `a2a-client/src/jsonrpc.rs` silently drop SSE events when **multiple complete events arrive in a single byte chunk and the byte stream ends immediately after**.

This causes callers to hang waiting for completion (e.g. terminal `Task` / `TASK_STATE_COMPLETED` `StatusUpdate`) even though the server has sent all events correctly.

## Root Cause

Both parsers use `stream::unfold` with state `(stream, buf)`. After parsing an event from the buffer, the closure returns early with `Some((Ok(event), (stream, buf)))`. On the next consumer poll, unfold re-enters and calls `stream.next().await` **before checking the buffer**. If the stream is already exhausted, `stream.next()` returns `None` and all remaining buffered events are silently dropped.

## Fix

Widen the unfold state to include a `VecDeque<Result<StreamResponse, A2AError>>` for pending events:

1. Parse every complete SSE event from `buf` into `pending` on each chunk
2. Drain `pending` one entry per consumer poll
3. Only call `stream.next().await` after `pending` is empty

Both `parse_sse_stream` (JSON-RPC envelope) and `parse_sse_stream_rest` (REST binding) share the same bug and are patched identically.

## Reproduction

A new test `test_parse_sse_stream_multiple_events_single_chunk` packs two complete SSE events into a single byte chunk and lets the stream end. The old implementation drops the second event; the fixed implementation delivers both.

Linuxdazhao:
## Summary

`parse_sse_stream` and `parse_sse_stream_rest` in `a2a-client/src/jsonrpc.rs` silently dropped every SSE event except the first one whenever **multiple complete events arrived in a single byte chunk and the byte stream ended immediately after** — a scenario that is routine in production (servers commonly flush a final burst of events right before closing the connection).

Symptom: streaming runs surfaced intermediate events but never the terminal `Task` / `TASK_STATE_COMPLETED` `StatusUpdate`, so callers hung waiting for completion despite the server having sent everything correctly. Reproduced consistently against a live server whose `curl` transcript showed all events delivered.

## Root cause

Both parsers drive their output via `stream::unfold` with state `(stream, buf)`:

```rust
loop {
    match stream.next().await {
        Some(Ok(chunk)) => {
            buf.push_str(...);
            while let Some(pos) = buf.find(\"\\n\\n\") {
                // …parse event…
                return Some((Ok(event), (stream, buf))); // ← exits after FIRST event
            }
        }
        None => return None,
        ...
    }
}
```

The early \`return\` inside the \`while let\` exits the closure after handing out one event. On the next consumer poll, unfold re-enters and **calls \`stream.next().await\` before looking at \`buf\`**. If the byte stream has already been exhausted, \`stream.next()\` returns \`None\`, the closure returns \`None\`, and every still-buffered event (including any terminal one) is silently dropped.

The existing test \`test_parse_sse_stream_multiple_events_separate_chunks\` did not catch this because it supplies each event as a **separate** byte chunk, which gives the parser a fresh \`stream.next()\` between events.

## Fix

Widen the unfold state to \`(stream, buf, pending: VecDeque<Result<StreamResponse, A2AError>>)\`:

- parse every complete SSE event from \`buf\` into \`pending\` on each chunk;
- drain \`pending\` one entry per consumer poll;
- only call \`stream.next().await\` after \`pending\` is empty.

Both variants (\`parse_sse_stream\` for the JSON-RPC envelope, \`parse_sse_stream_rest\` for the REST binding) are patched identically since they share the same bug.

## Tests

Two new tests added to \`a2a-client/src/jsonrpc.rs\`:

- \`test_parse_sse_stream_multiple_events_single_chunk\` (JSON-RPC variant)
- \`test_parse_sse_stream_rest_multiple_events_single_chunk\` (REST variant)

Both pack two complete SSE events into a **single** \`byte_stream\` chunk and then let the stream end, asserting both events reach the consumer. They fail on the previous implementation and pass with this change. The existing SSE test suite (18 tests in the \`jsonrpc::tests\` module) continues to pass.

## Test plan

- [x] \`cargo test -p a2a-client-lf --lib jsonrpc\` — 18/18 passing locally (including the 2 new regression tests)
- [x] CI green

---

Closes #52

### DIFF
--- a/a2a-client/src/jsonrpc.rs
+++ b/a2a-client/src/jsonrpc.rs
@@ -139,7 +139,7 @@ fn parse_sse_stream(
     let mapped = stream::unfold(
         (
             Box::pin(stream),
-            String::new(),
+            Vec::<u8>::new(),
             std::collections::VecDeque::<Result<StreamResponse, A2AError>>::new(),
         ),
         |(mut stream, mut buf, mut pending)| async move {
@@ -154,15 +154,24 @@ fn parse_sse_stream(
                 }
                 match stream.next().await {
                     Some(Ok(chunk)) => {
-                        buf.push_str(&String::from_utf8_lossy(&chunk));
+                        // Keep the buffer as raw bytes; defer UTF-8 decoding
+                        // until an event boundary is found. Per-chunk
+                        // `from_utf8_lossy` would silently replace multi-byte
+                        // characters split across chunk boundaries with
+                        // U+FFFD, corrupting user-visible text.
+                        buf.extend_from_slice(&chunk);
                         // Process complete SSE events (double newline terminated)
-                        while let Some(pos) = buf.find(
-                            "
-
-",
-                        ) {
-                            let event_text = buf[..pos].to_string();
-                            buf = buf[pos + 2..].to_string();
+                        while let Some(pos) = buf.windows(2).position(|w| w == b"\n\n") {
+                            let event_bytes: Vec<u8> = buf.drain(..pos + 2).collect();
+                            let event_text = match std::str::from_utf8(&event_bytes[..pos]) {
+                                Ok(s) => s,
+                                Err(e) => {
+                                    pending.push_back(Err(A2AError::internal(format!(
+                                        "SSE UTF-8 decode error: {e}"
+                                    ))));
+                                    continue;
+                                }
+                            };
 
                             // Extract data from SSE event
                             let mut data = String::new();
@@ -188,10 +197,8 @@ fn parse_sse_stream(
                             match serde_json::from_str::<JsonRpcResponse>(&data) {
                                 Ok(rpc_resp) => {
                                     if let Some(err) = rpc_resp.error {
-                                        pending.push_back(Err(A2AError::new(
-                                            err.code,
-                                            err.message,
-                                        )));
+                                        pending
+                                            .push_back(Err(A2AError::new(err.code, err.message)));
                                         continue;
                                     }
                                     if let Some(result) = rpc_resp.result {
@@ -217,9 +224,8 @@ fn parse_sse_stream(
                         // Loop back: pending may now hold events — drain before next read.
                     }
                     Some(Err(e)) => {
-                        pending.push_back(Err(A2AError::internal(format!(
-                            "SSE stream error: {e}"
-                        ))));
+                        pending
+                            .push_back(Err(A2AError::internal(format!("SSE stream error: {e}"))));
                     }
                     None => return None,
                 }
@@ -239,7 +245,7 @@ pub(crate) fn parse_sse_stream_rest(
     let mapped = stream::unfold(
         (
             Box::pin(stream),
-            String::new(),
+            Vec::<u8>::new(),
             std::collections::VecDeque::<Result<StreamResponse, A2AError>>::new(),
         ),
         |(mut stream, mut buf, mut pending)| async move {
@@ -252,14 +258,21 @@ pub(crate) fn parse_sse_stream_rest(
                 }
                 match stream.next().await {
                     Some(Ok(chunk)) => {
-                        buf.push_str(&String::from_utf8_lossy(&chunk));
-                        while let Some(pos) = buf.find(
-                            "
-
-",
-                        ) {
-                            let event_text = buf[..pos].to_string();
-                            buf = buf[pos + 2..].to_string();
+                        // Defer UTF-8 decoding until event boundary; see
+                        // parse_sse_stream for the rationale (per-chunk
+                        // lossy decode corrupts multi-byte chars).
+                        buf.extend_from_slice(&chunk);
+                        while let Some(pos) = buf.windows(2).position(|w| w == b"\n\n") {
+                            let event_bytes: Vec<u8> = buf.drain(..pos + 2).collect();
+                            let event_text = match std::str::from_utf8(&event_bytes[..pos]) {
+                                Ok(s) => s,
+                                Err(e) => {
+                                    pending.push_back(Err(A2AError::internal(format!(
+                                        "SSE UTF-8 decode error: {e}"
+                                    ))));
+                                    continue;
+                                }
+                            };
 
                             let mut data = String::new();
                             for line in event_text.lines() {
@@ -289,9 +302,8 @@ pub(crate) fn parse_sse_stream_rest(
                         }
                     }
                     Some(Err(e)) => {
-                        pending.push_back(Err(A2AError::internal(format!(
-                            "SSE stream error: {e}"
-                        ))));
+                        pending
+                            .push_back(Err(A2AError::internal(format!("SSE stream error: {e}"))));
                     }
                     None => return None,
                 }
@@ -803,8 +815,7 @@ mod tests {
         };
         let rpc_payload = |sr: StreamResponse, id: i64| -> String {
             let result = protojson_conv::to_value(&sr).unwrap();
-            serde_json::to_string(&JsonRpcResponse::success(JsonRpcId::Number(id), result))
-                .unwrap()
+            serde_json::to_string(&JsonRpcResponse::success(JsonRpcId::Number(id), result)).unwrap()
         };
         let combined = format!(
             "data: {}\n\ndata: {}\n\n",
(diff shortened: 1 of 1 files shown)


a2a-client/src/jsonrpc.rs
```diff
@@ -137,9 +137,21 @@ fn parse_sse_stream(
...
+                // Otherwise a single byte chunk carrying N complete SSE events
+                // would deliver only the first, and the remaining N-1 would be
+                // silently dropped if the connection closed before another
+                // chunk arrived (e.g. rapid final bursts with immediate close).
+                if let Some(item) = pending.pop_front() {
+                    return Some((item, (stream, buf, pending)));
+                }
                 match stream.next().await {
                     Some(Ok(chunk)) => {
                         buf.push_str(&String::from_utf8_lossy(&chunk));
```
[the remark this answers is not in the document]
Linuxdazhao: Addressed in the latest revision (aab1e87c): `buf` is now `Vec<u8>` and chunks are appended as raw bytes. UTF-8 decoding is deferred until a complete `\n\n`-delimited event is buffered, using strict `std::str::from_utf8` so split multi-byte characters are preserved and invalid UTF-8 surfaces as an `SSE UTF-8 decode error` instead of silent U+FFFD substitution. Regression tests cover split multi-byte, invalid UTF-8, and invalid protojson payloads.

a2a-client/src/jsonrpc.rs
```diff
@@ -234,9 +237,19 @@ pub(crate) fn parse_sse_stream_rest(
...
             loop {
+                // Drain already-parsed events before reading more bytes; see
+                // parse_sse_stream for the rationale (burst-then-close drops
+                // all but the first event without this).
+                if let Some(item) = pending.pop_front() {
+                    return Some((item, (stream, buf, pending)));
+                }
                 match stream.next().await {
                     Some(Ok(chunk)) => {
                         buf.push_str(&String::from_utf8_lossy(&chunk));
```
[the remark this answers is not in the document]
Linuxdazhao: Same fix applied here — see the reply on the `parse_sse_stream` thread. `parse_sse_stream_rest` uses the identical `Vec<u8>` buffer + deferred `std::str::from_utf8` pattern, so multi-byte characters split across chunk boundaries are preserved rather than replaced with U+FFFD.

a2a-client/src/jsonrpc.rs
```diff
@@ -234,19 +243,36 @@ pub(crate) fn parse_sse_stream_rest(
     stream: impl futures::Stream<Item = Result<bytes::Bytes, reqwest::Error>> + Send + 'static,
 ) -> BoxStream<'static, Result<StreamResponse, A2AError>> {
     let mapped = stream::unfold(
-        (Box::pin(stream), String::new()),
-        |(mut stream, mut buf)| async move {
+        (
+            Box::pin(stream),
+            Vec::<u8>::new(),
```
msardara: We can simplify this with `Vec::new()`

a2a-client/src/jsonrpc.rs
```diff
@@ -234,19 +243,36 @@ pub(crate) fn parse_sse_stream_rest(
...
+            std::collections::VecDeque::<Result<StreamResponse, A2AError>>::new(),
```
msardara: Let's import `std::collections::VecDeque` at the beginning of the module

```
use std::collections::VecDeque;
...
```

a2a-client/src/jsonrpc.rs
```diff
@@ -137,20 +137,41 @@ fn parse_sse_stream(
...
+                        buf.extend_from_slice(&chunk);
                         // Process complete SSE events (double newline terminated)
-                        while let Some(pos) = buf.find(
-                            "
-
-",
-                        ) {
-                            let event_text = buf[..pos].to_string();
-                            buf = buf[pos + 2..].to_string();
+                        while let Some(pos) = buf.windows(2).position(|w| w == b"\n\n") {
```
msardara: I was checking the SSE specifications at https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation

It looks like `\n` is not the only way we can terminate a line. `\r` and `\r\n` are possible options as well.

I'd fix the parser by handling the 3 possibilities with something in this line:

```rust
while let Some((start, end)) = find_event_boundary(&buf) {
    let event_bytes: Vec<u8> = buf.drain(..end).collect();
    // use buf[..start] as the event text

fn find_event_boundary(buf: &[u8]) -> Option<(usize, usize)> {
    for i in 0..buf.len().saturating_sub(1) {
        if buf[i] == b'\n' && buf[i + 1] == b'\n' {
            return Some((i, i + 2));
        }
        if buf[i] == b'\r' && buf[i + 1] == b'\r' {
            return Some((i, i + 2));
        }
        if i + 3 < buf.len() && &buf[i..i + 4] == b"\r\n\r\n" {
            return Some((i, i + 4));
        }
    }
    None
}
```

This is a pre-existing issue, but I think we can fix it now as we are touching this part, or create a follow-up PR to address this.

a2a-client/src/jsonrpc.rs
```diff
@@ -234,19 +243,36 @@ pub(crate) fn parse_sse_stream_rest(
...
-                            buf = buf[pos + 2..].to_string();
+                        // Defer UTF-8 decoding until event boundary; see
+                        // parse_sse_stream for the rationale (per-chunk
+                        // lossy decode corrupts multi-byte chars).
+                        buf.extend_from_slice(&chunk);
+                        while let Some(pos) = buf.windows(2).position(|w| w == b"\n\n") {
+                            let event_bytes: Vec<u8> = buf.drain(..pos + 2).collect();
+                            let event_text = match std::str::from_utf8(&event_bytes[..pos]) {
+                                Ok(s) => s,
+                                Err(e) => {
```
msardara: The functions parse_sse and parse_sse_rest are nearly identical (the only difference is how data is parsed (JSON-RPC envelope vs. direct StreamResponse)).

This was pre-existing as well, but I think we can simplify and have a single stream parse implementation.

```rust
fn parse_sse_bytes<F>(
      stream: impl Stream<...>,
      parse_event: F,
  ) -> BoxStream<...>
  where
      F: Fn(&str) -> Option<Result<StreamResponse, A2AError>> + Send + 'static,
```

We can simplify this in a follow-up PR.

Linuxdazhao:
Update on the remaining review feedback — all addressed in the latest push (`parse_sse_bytes` refactor + `\r`/`\r\n` support):

1. **SSE `\r\n` / `\r` line endings** — implemented `find_event_boundary()` that checks for `\n\n`, `\r\r`, and `\r\n\r\n` per the SSE spec. Added 4 unit tests for the helper + 2 integration tests (`test_parse_sse_stream_crlf_line_endings`, `test_parse_sse_stream_cr_line_endings`).

2. **Extracted `parse_sse_bytes<F>`** — both `parse_sse_stream` and `parse_sse_stream_rest` now delegate to this shared function, eliminating ~90 lines of duplicated parsing logic. Each caller passes its own `parse_event` closure (JSON-RPC envelope vs. direct StreamResponse).

3. **`VecDeque::new()`** — replaced `Default::default()` with explicit `VecDeque::new()` and added `use std::collections::VecDeque` at module top.

All 80 tests pass, clippy clean.

a2a-client/src/jsonrpc.rs
```diff
@@ -132,91 +133,79 @@ impl JsonRpcTransport {
...
-                                    }
-                                    data.push_str(d);
-                                } else if let Some(d) = line.strip_prefix("data:") {
-                                    if !data.is_empty() {
-                                        data.push('\n');
-                                    }
-                                    data.push_str(d);
+                        // Keep the buffer as raw bytes; defer UTF-8 decoding
+                        // until an event boundary is found. Per-chunk
+                        // `from_utf8_lossy` would silently replace multi-byte
```
msardara: This comment is referencing the fix of removing `from_utf8_lossy`. I'd keep these kind of comments in the PR / issue description rather than in the code itself.

a2a-client/src/jsonrpc.rs
```diff
@@ -768,6 +773,334 @@ mod tests {
...
+    /// Regression: multiple complete SSE events packed into a **single** byte
+    /// chunk must all be delivered, even when the upstream byte stream ends
+    /// immediately after that chunk.
+    ///
+    /// Prior to the drain-buffer fix, `parse_sse_stream` returned after parsing
+    /// the first event in the buffer and only re-polled the byte stream on the
+    /// next consumer call. If the byte stream was already exhausted (e.g. the
+    /// server closed the connection after a final burst of events), the
+    /// remaining events sitting in the parser buffer were silently dropped —
+    /// commonly losing the terminal `TASK_STATE_COMPLETED` `Task` event.
```
msardara: Also here, the comment references the fix itself. I'd just describe what this test is doing. The git history will tell us what the commit fixed.

msardara (commented):
I think the PR is in a good shape to be merged. Let's just fix the comments that are referencing this fix itself, and may confuse future readers as the old code will not be there after the PR is merged.

muscariello:
Where is the issue describing the problem this PR is solving ?

Every PR must have an issue.

Linuxdazhao:
Issue created and linked: https://github.com/a2aproject/a2a-rs/issues/52

Linuxdazhao:
@msardara comments referencing this fix have been cleaned up — pushed a new commit (`style(a2a-client): fix comment referencing old code in test doc`)

msardara (approved):
LGTM, thanks for the contribution @Linuxdazhao! ✅

Linuxdazhao:
Thanks for the review! Ready to merge when you have a chance.

msardara:
@Linuxdazhao the fix is available in the new crate: https://crates.io/crates/a2a-client-lf/0.1.15

PR #60, opened 2026-04-30
fix: allow disabling TLS in a2a-grpc

### LINKED ISSUE #59: Allow disabling TLS in a2a-grpc
### Problem

`a2a-grpc` depends on `a2a-client-lf` and `a2a-server-lf` with default features enabled, which unconditionally activates `rustls-tls-webpki-roots-no-provider` on reqwest. This forces all downstream consumers to install a rustls crypto provider at runtime, even when only using plain HTTP.

Applications that don't need TLS (e.g. test fixtures communicating over localhost) panic at startup with:

```
thread 'main' panicked at reqwest-0.12.28/src/async_impl/client.rs:767:29:
No provider set
```

### Proposed Solution

1. **`a2a-grpc/Cargo.toml`**: Declare `a2a-client` and `a2a-server` with `default-features = false`, and add explicit feature flags to forward TLS:

```toml
[dependencies]
a2a-client = { workspace = true, default-features = false }
a2a-server = { workspace = true, default-features = false }

[features]
default = ["rustls-tls"]
native-tls = ["a2a-client/native-tls", "a2a-server/native-tls"]
rustls-tls = ["a2a-client/rustls-tls", "a2a-server/rustls-tls"]
rustls = ["dep:tonic-tls", "dep:tokio-rustls", "rustls-tls"]
```

2. **`a2a-client/src`**: Gate build_reqwest_client_with_root_pem and with_root_certificates_pem methods behind #[cfg(any(feature = "rustls-tls", feature = "native-tls"))] so the crate compiles without TLS.

3. `**Workspace Cargo.toml`**: Set default-features = false on a2a-client and a2a-server workspace dependencies so member crates opt-in explicitly.

### Backward Compatibility

a2a-grpc defaults to ["rustls-tls"], so existing consumers see no behavior change. Only consumers that explicitly pass default-features = false get TLS-free builds.

msardara:
`a2a-grpc` depends on `a2a-client-lf` and `a2a-server-lf` with
default features enabled, which unconditionally activates
rustls-tls-webpki-roots-no-provider on reqwest.

This forces all downstream consumers to install a rustls crypto
provider at runtime, even when only using plain HTTP.

Fixes: #59

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,8 +20,8 @@ rust-version = "1.85"
 [workspace.dependencies]
 # Internal crates
 a2a = { package = "a2a-lf", path = "a2a", version = "0.2.4" }
-a2a-client = { package = "a2a-client-lf", path = "a2a-client", version = "0.1.14" }
-a2a-server = { package = "a2a-server-lf", path = "a2a-server", version = "0.2.7" }
+a2a-client = { package = "a2a-client-lf", path = "a2a-client", version = "0.1.14", default-features = false }
+a2a-server = { package = "a2a-server-lf", path = "a2a-server", version = "0.2.7", default-features = false }
 a2a-pb = { package = "a2a-pb", path = "a2a-pb", version = "0.1.7" }
 a2a-grpc = { package = "a2a-grpc", path = "a2a-grpc", version = "0.2.0" }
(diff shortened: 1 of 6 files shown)


micpapal (approved):
LGTM

PR #68, opened 2026-05-14
fix(a2a-client): unify reqwest TLS construction

### LINKED ISSUE #67: a2a-client: panic "No provider set" on first reqwest TLS path — missing rustls install_default()
## Summary

`a2a-client` (published as `a2a-client-lf` v0.1.16) panics with `No provider set` on the first reqwest TLS code path. Any consumer that builds an `A2AClient` over JSON-RPC, calls `AgentCardResolver::resolve`, or constructs `reqwest::Client::default()` for `JsonRpcTransport::new` triggers it — even when the target is plain HTTP, because reqwest 0.12 builds its rustls config eagerly at `Client::new` time.

Panic site:

```
thread 'a2a::sdk_client::tests::sdk_fetch_returns_local_card_shape'
panicked at .../reqwest-0.12.28/src/async_impl/client.rs:767:29:
No provider set
```

## Root cause

`a2a-client/Cargo.toml` selects the reqwest feature `rustls-tls-webpki-roots-no-provider`:

```toml
[features]
default = ["rustls-tls"]
rustls-tls = ["reqwest/rustls-tls-webpki-roots-no-provider", "dep:rustls"]

[dependencies.rustls]
version = "0.23"
features = ["aws-lc-rs"]
optional = true
```

The `-no-provider` reqwest feature requires the consuming code to register a rustls `CryptoProvider` *at runtime* via `CryptoProvider::install_default()` (or pass one explicitly to `ClientConfig`). Pulling `rustls = { features = ["aws-lc-rs"] }` into the build graph is not enough — `aws-lc-rs` has to be **installed as the process default** before the first `reqwest::Client` is built (rustls 0.23 design).

`a2a-server` and `a2a-grpc` already do this:

- `a2a-server/src/lib.rs`: `let _ = rustls::crypto::aws_lc_rs::default_provider().install_default();`
- `a2a-grpc/src/client.rs`: `let _ = tokio_rustls::rustls::crypto::aws_lc_rs::default_provider().install_default();`

`a2a-client/src/lib.rs` doesn't, so consumers panic the first time they touch the SDK.

## Reproduction

Any of these will panic on a default Tokio runtime:

```rust
use a2a_client::{agent_card::AgentCardResolver};

#[tokio::test]
async fn boom() {
    let _ = AgentCardResolver::new(None).resolve("http://127.0.0.1:0").await;
}
```

```rust
use a2a_client::jsonrpc::JsonRpcTransport;

#[tokio::test]
async fn boom_too() {
    let _t = JsonRpcTransport::new(reqwest::Client::default(), "http://x/a2a".into());
}
```

Confirmed with `a2a-client-lf 0.1.16` + reqwest `0.12.28` + rustls `0.23` (the versions resolved from crates.io for the default feature set).

## Proposed fix

Same one-liner already used in `a2a-server` and `a2a-grpc`, gated behind the `rustls-tls` feature, invoked from the shared reqwest client builder so all SDK entry points are covered:

```rust
// a2a-client/src/lib.rs
#[cfg(feature = "rustls-tls")]
pub(crate) fn ensure_default_crypto_provider() {
    static ONCE: std::sync::Once = std::sync::Once::new();
    ONCE.call_once(|| {
        let _ = rustls::crypto::aws_lc_rs::default_provider().install_default();
    });
}
```

Then call `ensure_default_crypto_provider()` at the head of `build_reqwest_client_with_root_pem` and at the entry of each public client constructor that ends up materializing a `reqwest::Client` (or once in `A2AClientFactory` / `AgentCardResolver::new`).

Happy to send a PR — would the project prefer (a) the `OnceLock`-guarded helper as above, or (b) just inlining `install_default()` at each call site to match the existing `a2a-server` / `a2a-grpc` style?

Linuxdazhao:
## Summary

`a2a-client`'s rustls path had two problems:

1. **Panic on first use.** Enabling `rustls-tls` pulled reqwest's
   `rustls-tls-webpki-roots-no-provider`, which requires the consumer to
   register a rustls `CryptoProvider` at runtime. Without an
   `install_default()` call, the first `reqwest::Client::new()` panicked
   with `"No provider set"` under `rustls 0.23+`. `a2a-server` and
   `a2a-grpc` already installed `aws-lc-rs`; `a2a-client` was missed.
2. **Silently dropped custom roots.** Custom CA bundles passed via
   `with_root_certificates_pem()` were ignored whenever a rustls
   provider feature was active, because the client was built via the
   plain `reqwest::Client::builder()` path that never saw them.

Rather than installing a process-global `CryptoProvider` (which leaks
state into the host application and races with other crates), this PR
unifies all internal `reqwest::Client` construction behind a single
constructor that wires TLS per-client.

Fixes: #67

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -42,6 +42,7 @@ dependencies = [
  "tokio-stream",
  "tracing",
  "uuid",
+ "webpki-roots",
 ]
 
 [[package]]
--- a/a2a-client/Cargo.toml
+++ b/a2a-client/Cargo.toml
@@ -1,3 +1,6 @@
+# Copyright AGNTCY Contributors (https://github.com/agntcy)
+# SPDX-License-Identifier: Apache-2.0
+
 [package]
 name = "a2a-client-lf"
 version = "0.1.16"
@@ -27,12 +30,33 @@ tracing = { workspace = true }
 uuid = { workspace = true }
 semver = { workspace = true }
 pin-project-lite = { workspace = true }
-rustls = { workspace = true, optional = true, features = ["aws-lc-rs"] }
+rustls = { workspace = true, optional = true }
+rustls-pemfile = { version = "2", optional = true }
+webpki-roots = { version = "1", optional = true }
 
 [dev-dependencies]
 rcgen = { workspace = true }
 
 [features]
-default = ["rustls-tls"]
+# By default we will use aws-lc-rs as crypto provider.
+default = ["rustls-tls-aws-lc-rs"]
+
+# For application setting their own provider, we use rustls
+# with NO crypto provider compiled in.
+# Consumers selecting only this feature are responsible for installing a
+# process-wide `CryptoProvider` (via `install_default()`) in their
+# main function. If they don't do it, `reqwest::Client::new()` will panic with
+# "no process-level CryptoProvider available".
+rustls-tls = [
+    "reqwest/rustls-tls-webpki-roots-no-provider",
+    "dep:rustls",
+    "dep:rustls-pemfile",
+    "dep:webpki-roots",
+]
+
+# If the consumer wants, a2a-client can also pull in a crypto provider and use it
+# when creating the reqwest client.
+rustls-tls-aws-lc-rs = ["rustls-tls", "rustls/aws-lc-rs"]
+rustls-tls-ring = ["rustls-tls", "rustls/ring"]
+
 native-tls = ["reqwest/native-tls"]
-rustls-tls = ["reqwest/rustls-tls-webpki-roots-no-provider", "dep:rustls"]
--- a/a2a-client/src/agent_card.rs
+++ b/a2a-client/src/agent_card.rs
@@ -11,7 +11,8 @@ pub struct AgentCardResolver {
 impl AgentCardResolver {
     pub fn new(client: Option<Client>) -> Self {
         AgentCardResolver {
-            client: client.unwrap_or_default(),
+            client: client
+                .unwrap_or_else(|| crate::default_reqwest_client(None).expect("default client")),
         }
     }
 
--- a/a2a-client/src/jsonrpc.rs
+++ b/a2a-client/src/jsonrpc.rs
@@ -43,6 +43,7 @@ pub struct JsonRpcTransport {
 }
 
 impl JsonRpcTransport {
+    /// Build a `JsonRpcTransport` from a pre-constructed `reqwest::Client`.
     pub fn new(client: Client, endpoint: String) -> Self {
         JsonRpcTransport { client, endpoint }
     }
@@ -458,14 +459,15 @@ pub struct JsonRpcTransportFactory {
 impl JsonRpcTransportFactory {
     pub fn new(client: Option<Client>) -> Self {
         JsonRpcTransportFactory {
-            client: client.unwrap_or_default(),
+            client: client
+                .unwrap_or_else(|| crate::default_reqwest_client(None).expect("default client")),
         }
     }
 
     #[cfg(any(feature = "rustls-tls", feature = "native-tls"))]
     pub fn with_root_certificates_pem(pem: &[u8]) -> Result<Self, A2AError> {
         Ok(Self {
-            client: crate::build_reqwest_client_with_root_pem(pem)?,
+            client: crate::default_reqwest_client(Some(pem))?,
         })
     }
 }
@@ -1014,7 +1016,10 @@ mod tests {
         let listener = TcpListener::bind("127.0.0.1:0").await.unwrap();
         let addr = listener.local_addr().unwrap();
         drop(listener);
-        let err = reqwest::get(format!("http://{addr}/"))
+        let err = crate::default_reqwest_client(None)
+            .unwrap()
+            .get(format!("http://{addr}/"))
+            .send()
             .await
             .expect_err("connection should fail");
 
@@ -1034,7 +1039,10 @@ mod tests {
         let listener = TcpListener::bind("127.0.0.1:0").await.unwrap();
         let addr = listener.local_addr().unwrap();
         drop(listener);
-        let err = reqwest::get(format!("http://{addr}/"))
+        let err = crate::default_reqwest_client(None)
+            .unwrap()
+            .get(format!("http://{addr}/"))
+            .send()
             .await
             .expect_err("connection should fail");
 
@@ -1117,7 +1125,10 @@ mod tests {
 
     #[test]
     fn test_jsonrpc_transport_new() {
-        let t = JsonRpcTransport::new(Client::new(), "http://localhost:8080".into());
+        let t = JsonRpcTransport::new(
+            crate::default_reqwest_client(None).unwrap(),
+            "http://localhost:8080".into(),
+        );
         assert_eq!(t.endpoint, "http://localhost:8080");
     }
 
@@ -1171,7 +1182,8 @@ mod tests {
         })
         .to_string();
         let (endpoint, request_rx) = spawn_jsonrpc_server(response).await;
-        let transport = JsonRpcTransport::new(Client::new(), endpoint);
+        let transport =
+            JsonRpcTransport::new(crate::default_reqwest_client(None).unwrap(), endpoint);
         let mut params = ServiceParams::new();
         params.insert("x-trace".into(), vec!["alpha".into(), "beta".into()]);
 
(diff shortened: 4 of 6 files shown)


msardara (changes requested):
Thanks for the PR @Linuxdazhao.

I think the choice of the crypto provider should be left to the consumer of the library, as it is a sensible choice of the application consuming the a2a-client crate. We should not set a process-wide crypto provider here.

The path I'd take is to expose a default feature flag that enables `aws-lc-rs` provider just for the reqwest client (so we don't set a process level one). In addition we should also give to consumers the possibility to opt-out and bring their own provider, by calling `install_default()` in the application (and not in a2a-client).

So, in a2a-client, we should set the rustls (and optionally webpki-roots) as optional dependenicies:

```toml
rustls = { workspace = true, optional = true }
webpki-roots = { version = "1", optional = true }
```

Then we should provide the following features:

```toml
[features]
# By defaut we will use aws-lc-rs as crypto provider
default = ["rustls-tls-aws-lc-rs"]

# For application setting their own provider, we use rustls
# with NO crypto provider compiled in.
# Consumers selecting only this feature are responsible for installing a
# process-wide `CryptoProvider` (via `install_default()`) in their
# main function. If they don't do it, `reqwest::Client::new()` will panic with
# "no process-level CryptoProvider available".
rustls-tls = [
    "reqwest/rustls-tls-webpki-roots-no-provider",
    "dep:rustls",
    "dep:webpki-roots",
]

# If the consumer wants, a2a-client can also pull in a crypto provider and use it
# when creating the reqwst client
rustls-tls-aws-lc-rs = ["rustls-tls", "rustls/aws-lc-rs"]
rustls-tls-ring = ["rustls-tls", "rustls/ring"]
```

Then in the code, when creating a default reqwest client, we use a configuration which is dictated by the feature flags:

```rust
/// Build a default `reqwest::Client` whose TLS layer matches this crate's
/// feature selection.
///
/// - When `rustls-tls-aws-lc-rs` or `rustls-tls-ring` is enabled, a
///   `rustls::ClientConfig` is constructed with the corresponding provider
///   and handed to `reqwest::ClientBuilder::use_preconfigured_tls`. No
///   process-global `CryptoProvider` is installed.
/// - When only `rustls-tls` is enabled (no provider variant), reqwest
///   falls back to whatever `CryptoProvider` the application installed via
///   `install_default()`. If none is installed, `reqwest::Client::new()`
///   panics; that is the contract the consumer accepted by disabling the
///   provider variants.
/// - When the rustls features are off (e.g. only `native-tls`), this just
///   returns the default reqwest client.
pub fn default_reqwest_client() -> reqwest::Client {
    let builder = reqwest::Client::builder();
    #[cfg(any(feature = "rustls-tls-aws-lc-rs", feature = "rustls-tls-ring"))]
    let builder = builder.use_preconfigured_tls(rustls_client_config());
    builder
        .build()
        .expect("failed to build default reqwest client")
}

#[cfg(any(feature = "rustls-tls-aws-lc-rs", feature = "rustls-tls-ring"))]
fn rustls_client_config() -> rustls::ClientConfig {
    let provider = std::sync::Arc::new(selected_crypto_provider());
    let mut roots = rustls::RootCertStore::empty();
    roots.extend(webpki_roots::TLS_SERVER_ROOTS.iter().cloned());
    rustls::ClientConfig::builder_with_provider(provider)
        .with_safe_default_protocol_versions()
        .expect("safe default protocol versions are supported")
        .with_root_certificates(roots)
        .with_no_client_auth()
}

#[cfg(feature = "rustls-tls-aws-lc-rs")]
fn selected_crypto_provider() -> rustls::crypto::CryptoProvider {
    rustls::crypto::aws_lc_rs::default_provider()
}

#[cfg(all(feature = "rustls-tls-ring", not(feature = "rustls-tls-aws-lc-rs")))]
fn selected_crypto_provider() -> rustls::crypto::CryptoProvider {
    rustls::crypto::ring::default_provider()
}
```

Linuxdazhao (commented):
Thanks for the detailed review and the concrete design! I've implemented exactly what you suggested:

- **Removed** `ensure_default_crypto_provider()` — no more process-wide `CryptoProvider`.
- **Added** `default_reqwest_client()` which uses `use_preconfigured_tls()` to pass a per-client `rustls::ClientConfig`, avoiding any global state.
- **Exposed** granular feature flags:
  - `default = ["rustls-tls-aws-lc-rs"]`
  - `rustls-tls` (no provider — consumer must call `install_default()`)
  - `rustls-tls-aws-lc-rs` (built-in `aws-lc-rs`)
  - `rustls-tls-ring` (built-in `ring`)
- **Moved** `rustls` and `webpki-roots` to optional dependencies.
- Updated `build_reqwest_client_with_root_pem()` to follow the same pattern.
- All 94 tests pass across `rustls-tls-aws-lc-rs`, `rustls-tls-ring`, `native-tls`, and bare `rustls-tls` feature combinations.

Pushed to the branch, ready for another look.

msardara (changes requested):
Thanks @Linuxdazhao! I tried it out using the examples and there is another issue to fix.

With the new constructor we are currently not taking into consideration the custom root certificate a client can provide when calling `RestTransportFactory::with_root_certificates_pem` or `JsonRpcTransportFactory::with_root_certificates_pem`.

We should unify the 2 constructors into a common helper that also takes care of creating a client config with that cert in the root store.

```rust
pub fn default_reqwest_client(
    extra_root_pem: Option<&[u8]>,
) -> Result<reqwest::Client, A2AError> {
    let builder = reqwest::Client::builder();

    #[cfg(any(feature = "rustls-tls-aws-lc-rs", feature = "rustls-tls-ring"))]
    let builder = builder.use_preconfigured_tls(rustls_client_config(extra_root_pem)?);

    #[cfg(not(any(feature = "rustls-tls-aws-lc-rs", feature = "rustls-tls-ring")))]
    let builder = match extra_root_pem {
        Some(pem) => {
            let cert = reqwest::Certificate::from_pem(pem)
                .map_err(|e| A2AError::internal(format!("invalid PEM certificate: {e}")))?;
            builder.add_root_certificate(cert)
        }
        None => builder,
    };

    builder
        .build()
        .map_err(|e| A2AError::internal(format!("failed to build HTTP client: {e}")))
}

#[cfg(any(feature = "rustls-tls-aws-lc-rs", feature = "rustls-tls-ring"))]
fn rustls_client_config(
    extra_root_pem: Option<&[u8]>,
) -> Result<rustls::ClientConfig, A2AError> {
    let provider = std::sync::Arc::new(selected_crypto_provider());
    let mut roots = rustls::RootCertStore::empty();
    roots.extend(webpki_roots::TLS_SERVER_ROOTS.iter().cloned());
    if let Some(pem) = extra_root_pem {
        for cert in rustls_pemfile::certs(&mut std::io::Cursor::new(pem)) {
            let der = cert
                .map_err(|e| A2AError::internal(format!("invalid PEM certificate: {e}")))?;
            roots
                .add(der)
                .map_err(|e| A2AError::internal(format!("failed to add CA: {e}")))?;
        }
    }
    Ok(rustls::ClientConfig::builder_with_provider(provider)
        .with_safe_default_protocol_versions()
        .expect("safe default protocol versions are supported")
        .with_root_certificates(roots)
        .with_no_client_auth())
}

#[cfg(feature = "rustls-tls-aws-lc-rs")]
fn selected_crypto_provider() -> rustls::crypto::CryptoProvider {
    rustls::crypto::aws_lc_rs::default_provider()
}

#[cfg(all(feature = "rustls-tls-ring", not(feature = "rustls-tls-aws-lc-rs")))]
fn selected_crypto_provider() -> rustls::crypto::CryptoProvider {
    rustls::crypto::ring::default_provider()
}
```

And then when creating a client you simplu call 

```rust
crate::default_reqwest_client(Some(pem))
```

or

```rust
crate::default_reqwest_client(None)
```

We can also remove the `build_reqwest_client_with_root_pem` helper, as now we have a unified one taking care of all the cases.

Linuxdazhao:
@msardara Addressed the review feedback. I've unified `default_reqwest_client()` and `build_reqwest_client_with_root_pem()` into a single `default_reqwest_client(extra_root_pem: Option<&[u8]>)` constructor.

**Key changes:**

1. **Unified constructor** - `default_reqwest_client` now accepts an optional PEM bundle
2. **Rustls path** - Custom certificates are properly added to the rustls root store via `rustls_pemfile::certs()` alongside webpki roots
3. **Non-rustls path** - Uses `reqwest::Certificate::from_pem()` + `add_root_certificate()`
4. **Removed** `build_reqwest_client_with_root_pem()` (no longer needed)
5. **Added** `rustls-pemfile` dependency gated behind `rustls-tls` feature

The `with_root_certificates_pem()` factories now simply call:
```rust
crate::default_reqwest_client(Some(pem))?
```

And the default case:
```rust
crate::default_reqwest_client(None).expect("default client")
```

This should fix the bug where custom root certificates were silently ignored when using `rustls-tls-aws-lc-rs` or `rustls-tls-ring` features.

msardara (approved):
LGTM, thanks @Linuxdazhao .

I also updated the PR description accordingly to the changes.

Issue #69, opened 2026-05-19
[Bug]: ServiceParams is never populated by the REST or JSON-RPC HTTP transports


justintime4tea:
## Summary

The axum-based HTTP transport in `a2a-server` constructs an empty `ServiceParams` at every entry point and passes it through to the `RequestHandler`. As a result, downstream `AgentExecutor` implementations always see `ctx.service_params` as an empty map, regardless of what headers the HTTP client sent. This breaks any executor logic that depends on transport-level metadata (e.g. forwarding an incoming `Authorization` header to downstream services, request correlation IDs).

The gRPC and slimrpc transport servers in populate `ServiceParams` correctly from their respective transport metadata, so the design intent is clear — the HTTP transport just doesn't have the equivalent extractor.

## Affected version

Reproduced against `main` at commit `73c72ee` **(v0.3.0?)**

Crates affected: `a2a-server` (specifically `rest.rs` and `jsonrpc.rs`).

## Reproduction

Minimal `AgentExecutor` that prints whatever it receives:

```rust
struct DebugExecutor;

#[async_trait::async_trait]
impl AgentExecutor for DebugExecutor {
    fn execute(&self, ctx: ExecutorContext)
        -> BoxStream<'static, Result<StreamResponse, A2AError>>
    {
        eprintln!("service_params = {:?}", ctx.service_params);
        eprintln!("user           = {:?}", ctx.user);
        // ... return a trivial completed task
    }

    fn cancel(&self, ctx: ExecutorContext) -> BoxStream<'static, Result<StreamResponse, A2AError>> {
        Box::pin(futures::stream::empty())
    }
}
```

Mount the REST router:

```rust
let handler = Arc::new(DefaultRequestHandler::new(DebugExecutor, InMemoryTaskStore::new()));
let app = Router::new().nest("/a2a/v1", rest::rest_router(handler));
```

Send a request with arbitrary headers:

```bash
curl -X POST http://localhost:8080/a2a/v1/message:send \
-H 'Authorization: Bearer test-token-abc' \
-H 'X-Tenant-Id: acme' \
-H 'Content-Type: application/json' \
-d '{"message":{"messageId":"m1","role":"ROLE_USER","parts":[{"text":"hi"}]}}'
```

**Observed output:**

```
service_params = {}
user           = None
```

The same setup behind `a2a-grpc` returns a populated `service_params` map containing all incoming gRPC metadata.

## Expected behavior

[...]

PR #70, opened 2026-05-19
fix(a2a-server): propagate HTTP headers into ServiceParams

justintime4tea:
## Summary

The axum-based HTTP (`rest`) and JSON-RPC transports in `a2a-server` constructed an empty `ServiceParams` at every entry point, so downstream `AgentExecutor` and `RequestHandler` implementations always saw `ctx.service_params` as `{}` regardless of what headers the client sent. This broke any executor or handler logic that depends on transport metadata — forwarding an incoming `Authorization` header to a downstream service, tenant routing, request-correlation IDs, etc.

The gRPC and slimrpc transports in this workspace already populate `ServiceParams` from their transport metadata (`a2a-grpc/src/server.rs:34`, `a2a-slimrpc/src/common.rs:70`). This change brings the HTTP transports to parity.

This should resolve issue #69 

## Changes

- **`a2a-server/src/rest.rs`** — added `extract_service_params(&HeaderMap) -> ServiceParams` and wired a `HeaderMap` extractor through every axum handler (`handle_send_message`, `handle_stream_message`, `handle_get_task_or_subscribe`, `handle_list_tasks`, `handle_post_task_action`, `handle_cancel_task_legacy`, `handle_subscribe_to_task_legacy`, `handle_create_push_config`, `handle_get_push_config`, `handle_list_push_configs`, `handle_delete_push_config`, `handle_get_extended_agent_card`). The private `_inner` helpers now take a `ServiceParams` argument from their callers.
- **`a2a-server/src/jsonrpc.rs`** — added the same helper and wired a `HeaderMap` extractor through `handle_jsonrpc`.

No public API signatures changed. `rest_router`, `jsonrpc_router`, `RestState`, `JsonRpcState`, and the `RequestHandler` trait are all untouched.

## Behavioral note: incoming request headers now reach executors

This change **exposes incoming HTTP request headers to executor / handler code** via `ctx.service_params`. This includes potentially sensitive headers like `Authorization`, `Cookie`, and `X-Api-Key`.

This is **the same exposure model the gRPC and slimrpc transports already provide** — both surface inbound transport metadata (gRPC metadata / slim context metadata) into the same `ServiceParams` map. The HTTP transports were the outlier, and the doc comment on `ServiceParams` always advertised this contract:

```rust
// a2a-server/src/middleware.rs
/// Service parameters — metadata from HTTP headers or gRPC metadata.
pub type ServiceParams = HashMap<String, Vec<String>>;
```

Conventions to be aware of:

- **Header names are lowercased** (axum normalizes them, matching HTTP/2 semantics). Look up headers as `params.get("authorization")`, not `"Authorization"`.
- **Multi-valued headers** produce `Vec<String>` entries in insertion order.
- **Non-ASCII / opaque-binary header values** are silently skipped, mirroring gRPC's handling of binary metadata entries.
- **No header filtering.** Hop-by-hop and framing headers (`host`, `content-length`, `accept`, `user-agent`, etc.) flow through verbatim. This matches gRPC, which also does not filter pseudo-headers.

**Operator guidance:** if your executor or `CallInterceptor` logs `ctx.service_params` wholesale (e.g. for debugging), audit those log statements before deploying this version — auth tokens and other secrets will now appear in those logs. The same caveat already applies to deployments using the gRPC or slimrpc transports.

## Test plan

A failing test for each transport was added first to reproduce the bug, then the fix was applied. Coverage spans every endpoint, not just the executor-driven ones:

### REST (`a2a-server/src/rest.rs`)

Executor-driven endpoints (via `CapturingExecutor` recording `ctx.service_params`):
- [x] `test_send_message_propagates_headers_as_service_params`
- [x] `test_stream_message_propagates_headers_as_service_params`

Non-executor endpoints (via a new `CapturingHandler` stub recording `ServiceParams` per `RequestHandler` method):
- [x] `test_get_task_propagates_headers_as_service_params`
- [x] `test_list_tasks_propagates_headers_as_service_params`
- [x] `test_cancel_task_propagates_headers_as_service_params`
- [x] `test_subscribe_to_task_propagates_headers_as_service_params`
- [x] `test_create_push_config_propagates_headers_as_service_params`
- [x] `test_get_push_config_propagates_headers_as_service_params`
- [x] `test_list_push_configs_propagates_headers_as_service_params`
- [x] `test_delete_push_config_propagates_headers_as_service_params`
- [x] `test_get_extended_agent_card_propagates_headers_as_service_params`

### JSON-RPC (`a2a-server/src/jsonrpc.rs`)

- [x] `test_send_message_propagates_headers_as_service_params` (unary)
- [x] `test_streaming_send_message_propagates_headers_as_service_params` (SSE)

### Verification

- [x] All 117 unit tests in `a2a-server` pass (the 13 propagation tests above + 104 pre-existing).
- [x] `cargo test --workspace` green.
- [x] `cargo clippy -p a2a-server-lf --all-targets` clean.

## Out of scope

- `ctx.user` remains hardcoded to `None` in `DefaultRequestHandler`. Surfacing an authenticated user from an interceptor hook is a separate, larger design choice and is intentionally deferred from this fix.
- No header deny-list / filtering layer is added; behavior matches the gRPC transport's pass-through model. Operators who want to redact specific headers can do so via a `CallInterceptor` or an axum middleware layer.

### DIFF
--- a/a2a-server/src/lib.rs
+++ b/a2a-server/src/lib.rs
@@ -21,10 +21,175 @@ pub use task_store::{InMemoryTaskStore, TaskStore};
 
 #[cfg(test)]
 pub(crate) mod test_util {
+    use std::collections::HashMap;
+    use std::sync::{Arc, Mutex};
+
+    use a2a::*;
+    use async_trait::async_trait;
+    use futures::stream::BoxStream;
+
+    use crate::handler::RequestHandler;
+    use crate::middleware::ServiceParams;
+
     pub fn install_crypto_provider() {
         #[cfg(feature = "rustls-tls")]
         {
             let _ = rustls::crypto::aws_lc_rs::default_provider().install_default();
         }
+    }
+
+    pub(crate) type CapturedParams = Arc<Mutex<HashMap<&'static str, ServiceParams>>>;
+
+    /// `RequestHandler` stub that records the `ServiceParams` it sees per method.
+    ///
+    /// Used by transport-layer tests to assert that incoming request headers
+    /// are surfaced through to the `RequestHandler` boundary — which is exactly
+    /// what the HTTP / JSON-RPC transports are responsible for. Returns minimal
+    /// stub responses so tests don't depend on `DefaultRequestHandler` behavior.
+    pub(crate) struct CapturingHandler {
+        captured: CapturedParams,
+    }
+
+    impl CapturingHandler {
+        pub fn new() -> (Arc<Self>, CapturedParams) {
+            let captured = Arc::new(Mutex::new(HashMap::new()));
+            let handler = Arc::new(CapturingHandler {
+                captured: captured.clone(),
+            });
+            (handler, captured)
+        }
+
+        fn record(&self, method: &'static str, params: &ServiceParams) {
+            self.captured.lock().unwrap().insert(method, params.clone());
+        }
+    }
+
+    #[async_trait]
+    impl RequestHandler for CapturingHandler {
+        async fn send_message(
+            &self,
+            params: &ServiceParams,
+            _req: SendMessageRequest,
+        ) -> Result<SendMessageResponse, A2AError> {
+            self.record("send_message", params);
+            Err(A2AError::internal("stub"))
+        }
+
+        async fn send_streaming_message(
+            &self,
+            params: &ServiceParams,
+            _req: SendMessageRequest,
+        ) -> Result<BoxStream<'static, Result<StreamResponse, A2AError>>, A2AError> {
+            self.record("send_streaming_message", params);
+            Ok(Box::pin(futures::stream::empty()))
+        }
+
+        async fn get_task(
+            &self,
+            params: &ServiceParams,
+            _req: GetTaskRequest,
+        ) -> Result<Task, A2AError> {
+            self.record("get_task", params);
+            Err(A2AError::task_not_found("stub"))
+        }
+
+        async fn list_tasks(
+            &self,
+            params: &ServiceParams,
+            _req: ListTasksRequest,
+        ) -> Result<ListTasksResponse, A2AError> {
+            self.record("list_tasks", params);
+            Ok(ListTasksResponse {
+                tasks: Vec::new(),
+                next_page_token: String::new(),
+                page_size: 0,
+                total_size: 0,
+            })
+        }
+
+        async fn cancel_task(
+            &self,
+            params: &ServiceParams,
+            _req: CancelTaskRequest,
+        ) -> Result<Task, A2AError> {
+            self.record("cancel_task", params);
+            Err(A2AError::task_not_found("stub"))
+        }
+
+        async fn subscribe_to_task(
+            &self,
+            params: &ServiceParams,
+            _req: SubscribeToTaskRequest,
+        ) -> Result<BoxStream<'static, Result<StreamResponse, A2AError>>, A2AError> {
+            self.record("subscribe_to_task", params);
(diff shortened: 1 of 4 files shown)


a2a-server/src/lib.rs
```diff
@@ -21,10 +21,175 @@ pub use task_store::{InMemoryTaskStore, TaskStore};
...
+    use crate::middleware::ServiceParams;
+
     pub fn install_crypto_provider() {
         #[cfg(feature = "rustls-tls")]
         {
             let _ = rustls::crypto::aws_lc_rs::default_provider().install_default();
         }
     }
+
+    pub type CapturedParams = Arc<Mutex<HashMap<&'static str, ServiceParams>>>;
```
msardara:
```suggestion
    pub(crate) type CapturedParams = Arc<Mutex<HashMap<&'static str, ServiceParams>>>;
```

a2a-server/src/lib.rs
```diff
@@ -21,10 +21,175 @@ pub use task_store::{InMemoryTaskStore, TaskStore};
...
+
+    /// `RequestHandler` stub that records the `ServiceParams` it sees per method.
+    ///
+    /// Used by transport-layer tests to assert that incoming request headers
+    /// are surfaced through to the `RequestHandler` boundary — which is exactly
+    /// what the HTTP / JSON-RPC transports are responsible for. Returns minimal
+    /// stub responses so tests don't depend on `DefaultRequestHandler` behavior.
+    pub struct CapturingHandler {
```
msardara:
```suggestion
    pub(crate) struct CapturingHandler {
```

a2a-server/src/lib.rs
```diff
@@ -21,10 +21,175 @@ pub use task_store::{InMemoryTaskStore, TaskStore};
...
+            _req: GetExtendedAgentCardRequest,
+        ) -> Result<AgentCard, A2AError> {
+            self.record("get_extended_agent_card", params);
+            Err(A2AError::internal("stub"))
+        }
+    }
+
+    /// Assert that `method` was called with a service-params entry matching `header = expected`.
+    #[track_caller]
+    pub fn assert_header_captured(
```
msardara:
```suggestion
    pub(crate) fn assert_header_captured(
```

a2a-server/src/middleware.rs
```diff
@@ -2,12 +2,33 @@
...
+    for name in headers.keys() {
+        let values: Vec<String> = headers
+            .get_all(name)
+            .iter()
+            .filter_map(|v| v.to_str().ok().map(str::to_owned))
+            .collect();
+        if !values.is_empty() {
+            params.insert(name.as_str().to_owned(), values);
+        }
+    }
```
msardara:
```suggestion
    for (name, value) in headers {
        if let Ok(v) = value.to_str() {
            params
                .entry(name.as_str().to_owned())
                .or_default()
                .push(v.to_owned());
        }
    }
```

msardara (changes requested):
Thanks for the PR @justintime4tea! LGTM, just few nits to fix before merging.

justintime4tea:
> Thanks for the PR @justintime4tea! LGTM, just few nits to fix before merging.

Thanks @msardara! I think I resolved things, could you give it another look?

msardara (approved):

msardara:
@justintime4tea fixed in [a2a-server-lf 0.3.1](https://crates.io/crates/a2a-server-lf) ✅ 

Thanks a lot for the contribution!

PR #79, opened 2026-06-10
fix(a2a-client): surface plain JSON-RPC error responses to streaming calls

arkavo-com:
## Problem

`parse_sse_bytes` discards any unterminated buffer at end-of-stream. A server that answers a streaming call (`SendStreamingMessage`, `SubscribeToTask`) with a plain JSON-RPC error envelope — `Content-Type: application/json`, no SSE framing — therefore produces an **empty, successfully-closed stream**, and the caller never sees the error. `a2a-server` itself responds this way for pre-stream errors (e.g. `-32004` from `SubscribeToTask` on a terminal task), so the SDK's own client+server pair silently loses these errors today.

## Fix

Track whether any SSE event boundary was observed; at end-of-stream with no SSE framing and a non-empty residual buffer, parse the buffer as a `JsonRpcResponse` envelope. A contained error is emitted through the same mapping as SSE-framed envelope errors; a contained result is delivered as a normal `StreamResponse`; anything else keeps the existing clean-end behavior. SSE streams with trailing partial events are unaffected.

## Found by / verified with

Cross-SDK interop testing ([a2a-conformance](https://github.com/arkavo-ai/a2a-conformance), scenario `errors/unsupported-operation-32004`; [FINDINGS.md](https://github.com/arkavo-ai/a2a-conformance/blob/main/FINDINGS.md) finding 3): the a2a-rs client vs. its own server previously read the `-32004` envelope as an empty stream; with this branch the scenario passes. 5 new unit tests; `cargo test --workspace` 452 passed / 0 failed (baseline on main: 447/0); fmt + clippy clean.

## Related spec note

The underlying ambiguity isn't this SDK's fault: §9.4.2 doesn't specify whether a pre-stream protocol error is delivered as plain JSON or as an SSE-framed event, and implementations have picked both. Happy to open a spec issue on a2aproject/A2A if maintainers think it's worth pinning down — this fix makes the client robust to either choice.

### DIFF
--- a/a2a-client/src/jsonrpc.rs
+++ b/a2a-client/src/jsonrpc.rs
@@ -172,6 +172,27 @@ fn find_event_boundary(buf: &[u8]) -> Option<(usize, usize)> {
     None
 }
 
+/// Interpret a residual, non-SSE-framed response body as a plain JSON-RPC
+/// response envelope.
+///
+/// Some servers answer a streaming call that fails before any stream is
+/// produced (e.g. `-32004 UnsupportedOperation`) with a plain
+/// `application/json` JSON-RPC envelope instead of SSE framing. Such a body
+/// contains no event boundary, so without this fallback it would be
+/// discarded at end-of-stream and the caller would see an empty,
+/// successfully-closed stream instead of the error.
+fn parse_plain_json_tail(buf: &[u8]) -> Option<Result<StreamResponse, A2AError>> {
+    let rpc_resp: JsonRpcResponse = serde_json::from_slice(buf).ok()?;
+    if let Some(err) = rpc_resp.error {
+        return Some(Err(parse_jsonrpc_error(err)));
+    }
+    let result = rpc_resp.result?;
+    match protojson_conv::from_value::<StreamResponse>(result) {
+        Ok(sr) => Some(Ok(sr)),
+        Err(e) => Some(Err(A2AError::internal(format!("SSE parse error: {e}")))),
+    }
+}
+
 /// Shared SSE byte-stream parser that abstracts over the event-parse callback.
 ///
 /// `parse_event` is invoked on each complete `\n\n`/`\r\r`/`\r\n\r\n`-delimited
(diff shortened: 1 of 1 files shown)


muscariello:
already done #198

muscariello:
Thanks for this, @arkavo-com — and for the cross-SDK interop testing that found it. The defect was real and your diagnosis of it was correct: `parse_sse_bytes` discarding its residual buffer at end-of-stream turned a cut stream into an empty, successfully-closed one, and the SDK's own client/server pair lost `-32004` because of it.

Closing in favour of #198, which carries the fix forward. Tracked as #197.

Two things changed since June that made a fresh patch cleaner than a rebase of this one:

**Part of it landed by another route.** `send_streaming_message` gained a `Content-Type` check on `main`, so an `application/json` body is now parsed as a `JsonRpcResponse` and its error surfaced before the SSE parser ever sees it. That covers `a2a-server`'s own pre-stream errors, which set the header. The REST binding turned out never to be affected: `post_streaming`/`get_streaming` check `resp.status().is_success()` first, and a REST error carries a non-2xx status.

**What remained was narrower, and partly different.** Three cases still ended silently, and the third is not one this PR addressed:

1. a response with **no `Content-Type`** — that check deliberately assumes a header-less response is a stream, since inspecting the body would hang one that has not sent anything yet;
2. a response declaring **`text/event-stream`** that carries a plain envelope anyway;
3. a **truncated final SSE event** — a server that writes `data: {...}` and closes without the trailing blank line. This one loses *data*, not just errors.

Two places where #198 departs from the approach here, both because of (3):

- The tail is handed to the same `parse_event` callback the framed path uses, rather than being parsed as a `JsonRpcResponse`. The REST binding shares this parser but its events are bare `StreamResponse` JSON, so an envelope-only parse cannot deliver a truncated REST event.
- A tail that parses as neither an event nor an envelope is reported rather than returning a clean end. A non-empty unparseable remainder means the connection was cut mid-message, and ending silently is the one answer that is always wrong.

Your `.fuse()` is kept — the unfold polls once more after emitting the tail, and polling a stream after it has returned `None` is allowed to panic. That was easy to miss and would have been a latent panic.

The spec point you raised is still open and still worth pinning down: §9.4.2 does not say whether a pre-stream protocol error is delivered as plain JSON or as an SSE-framed event, and implementations have picked both. #198 makes the client robust to either, but an upstream issue on a2aproject/A2A would be welcome if you still want to file it.

PR #80, opened 2026-06-10
fix(a2a-client): echo selected interface tenant on requests (spec §8.3.2)

arkavo-com:
## Problem

Spec §8.3.2 rule 4: clients MUST "set the `tenant` field in every request message to exactly the value declared in the selected `AgentInterface` entry (omit the field if `tenant` is not set in that entry)". `a2a-client` has no mechanism for this — request types carry `tenant: Option<String>` but nothing populates it from the agent card, so multi-tenant A2A deployments cannot be addressed through this SDK.

## Fix

- `A2AClient` gains `with_tenant(impl Into<String>)`, a `tenant()` getter, and a `Cow`-based `apply_tenant` helper applied in all 11 request methods (including `get_extended_agent_card`). An explicitly set request tenant always wins; with no client tenant there is no clone and no behavior change.
- `A2AClientFactory::create_from_card` sets the tenant from the selected interface, treating the empty string as undeclared (proto3 default semantics).

No changes to `a2a` types were needed (request types already derive `Clone`).

**Semver note:** adds public API (`with_tenant`, `tenant()`) — a minor-version addition. Behavior is opt-in: nothing changes unless a tenant is configured or the selected card interface declares one.

## Found by / verified with

Cross-SDK interop testing ([a2a-conformance](https://github.com/arkavo-ai/a2a-conformance), scenario `discovery/tenant-echo` with the runner observing the request server-side; [FINDINGS.md](https://github.com/arkavo-ai/a2a-conformance/blob/main/FINDINGS.md) finding 1): previously failed against every server including a2a-rs's own; passes with this branch. 6 new unit tests (all-methods propagation, explicit-tenant-wins, factory propagation incl. empty-string-as-unset); `cargo test --workspace` 453 passed / 0 failed (baseline 447/0); fmt + clippy clean.

### DIFF
--- a/a2a-client/src/client.rs
+++ b/a2a-client/src/client.rs
@@ -4,6 +4,7 @@ use a2a::jsonrpc::methods;
 use a2a::*;
 use async_trait::async_trait;
 use futures::stream::BoxStream;
+use std::borrow::Cow;
 use std::sync::Arc;
 
 use crate::middleware::CallInterceptor;
@@ -14,6 +15,7 @@ pub struct A2AClient<T: Transport> {
     transport: T,
     interceptors: Vec<Arc<dyn CallInterceptor>>,
     default_params: ServiceParams,
+    tenant: Option<String>,
 }
 
 impl<T: Transport> A2AClient<T> {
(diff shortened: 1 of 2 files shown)


muscariello:
superseed by #200

PR #81, opened 2026-06-10
fix(a2a-pb): ignore unknown fields in ProtoJSON decode (spec §5.7)

arkavo-com:
## Problem

Spec §5.7: "Implementations SHOULD ignore unrecognized fields in messages, allowing for forward compatibility as the protocol evolves." The pbjson-generated deserializers reject unknown fields, so any peer that sends a forward-compatible extra field (or a newer protocol revision) fails to decode.

## Fix

One line in `a2a-pb/build.rs`: `.ignore_unknown_fields()` on the `pbjson_build::Builder` (supported by the pbjson-build 0.9.0 already in the workspace). Unknown fields now route to the generated `__SkipField__` arm. All existing post-generation replacements (`NULL_REPEATED_FIELD_REPLACEMENTS`, map handling) still match — the generated serde code lives in `OUT_DIR`, so no committed generated files change.

`ignore_unknown_enum_variants` is deliberately **not** enabled: unknown enum *string values* still error, preserving strictness where the spec doesn't ask for leniency.

## Heads-up for review: one existing test fixture changed

`a2a-server`'s `test_invalid_params` used `{"bogus": true}` — a merely-unknown field — as its invalid-params fixture. Under §5.7 semantics that payload is now *valid* (decodes to a default request). The test now uses a genuine type mismatch (`{"message": "not-an-object"}`) with the same error-code assertion, and a new `test_unknown_params_fields_are_ignored` pins the tolerance end-to-end through the JSON-RPC server.

## Found by / verified with

Cross-SDK interop testing ([a2a-conformance](https://github.com/arkavo-ai/a2a-conformance), scenario `edge/unknown-field-tolerance`; [FINDINGS.md](https://github.com/arkavo-ai/a2a-conformance/blob/main/FINDINGS.md) finding 2). New round-trip tests in `a2a-pb/tests/` (unknown top-level + nested fields accepted; unknown enum values still rejected); `cargo test --workspace` 450 passed / 0 failed (baseline 447/0); fmt + clippy clean.

### DIFF
--- a/a2a-pb/build.rs
+++ b/a2a-pb/build.rs
@@ -120,6 +120,10 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     let mut builder = pbjson_build::Builder::new();
     builder.out_dir(&out_dir);
     builder.register_descriptors(&descriptor_set)?;
+    // Spec §5.7: implementations SHOULD ignore unrecognized fields in
+    // messages, allowing for forward compatibility. Unknown enum string
+    // values are still rejected.
+    builder.ignore_unknown_fields();
     builder.build(PROTOJSON_PACKAGES)?;
     patch_generated_protojson_serde(&out_dir)?;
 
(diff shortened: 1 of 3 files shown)


muscariello (approved):
LGTM

PR #86, opened 2026-06-12
fix(itk): serve agent card at HTTP root for ITK readiness check

muscariello:
## Summary

The ITK service (`testlib.py`) polls `GET /.well-known/agent-card.json` on the
agent's HTTP port to verify readiness before running any tests. The upstream
`agents/rust/v10/src/main.rs` only mounts `agent_card_router` under the
`/jsonrpc` and `/rest` nested sub-routers, so the root-level poll returns 404
and the readiness check times out regardless of whether the agent is actually up.

Add `itk/src/main.rs` derived from the upstream source with one fix: also merge
`agent_card_router` at the root of `http_app`:

```rust
let http_app = Router::new()
    .merge(a2a_server::agent_card::agent_card_router(card_producer))  // ← root
    .nest("/jsonrpc", jsonrpc_sub)
    .nest("/rest", rest_sub)
    ...
```

Update `[[bin]] path` in `itk/Cargo.toml` to point to `src/main.rs`.

Fixes: https://github.com/a2aproject/a2a-rs/actions/runs/27401760519

## Test plan

- [ ] Trigger `workflow_dispatch` on `itk-nightly` and verify readiness check passes
- [ ] Confirm `itk_rust.json` has non-empty `scenarios`


muscariello:
Fix moved to upstream repo: https://github.com/a2aproject/a2a-itk/pull/11

PR #95, opened 2026-06-26
fix(winget): use WINGET_PAT for cross-org PR to microsoft/winget-pkgs

muscariello:
GitHub App tokens are scoped to the org they're installed in and cannot open PRs on `microsoft/winget-pkgs`.

## Changes
- Add `WINGET_PAT` env var support: a classic PAT with `public_repo` scope, used only for the `gh pr create` step against `microsoft/winget-pkgs`
- Fork push still uses the App token (which has write access to `a2aproject/winget-pkgs`)
- If `WINGET_PAT` is unset, manifests are pushed to the fork but PR submission is skipped with a `::notice::` instead of failing

## Setup required
Add a `WINGET_PAT` repository secret: a classic PAT from any org member with `public_repo` scope.

### DIFF
--- a/.github/workflows/winget-manifests.yml
+++ b/.github/workflows/winget-manifests.yml
@@ -75,6 +75,7 @@ jobs:
       - name: Open or update winget-pkgs PR
         env:
           WINGET_PKGS_TOKEN: ${{ steps.app-token.outputs.token }}
+          WINGET_PAT: ${{ secrets.WINGET_PAT }}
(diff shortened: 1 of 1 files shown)


micpapal (approved):

muscariello:
Closing — root cause is a missing Pull requests: Read & Write permission on the GitHub App, not a need for a PAT. The App token approach (same as shadi) is correct.

PR #111, opened 2026-07-22
feat(slimrpc): add Collaborate — many-to-many messaging on SLIM group channels

muscariello:
## Summary

Implements the demo-sufficient subset of the [SLIMRPC collaborative channel extension](https://github.com/a2aproject/experimental-cpb-slimrpc/pull/5) (currently a draft spec, open for review):

```protobuf
rpc Collaborate(stream Message) returns (stream Message);
```

Unlike the existing multicast `SendMessage`/`SendStreamingMessage` (one client → many independent agent responses), `Collaborate` is genuinely many-to-many: any group-channel member may originate messages, and every message is broadcast to every other member. Reuses A2A's own `Message` type in both directions — no new proto types.

## Changes

- `SlimRpcTransport::new_group`/`new_group_with_connection` — build a transport over a SLIM **group** channel (`slim_rpc::Channel::new_group`), as opposed to the existing point-to-point `new`/`new_with_connection`.
- `SlimRpcTransport::collaborate(outbound, timeout)` — drives `Channel::multicast_stream_stream` with A2A `Message` protos, and stamps every inbound `Message` with `metadata["slim-src"]` = the sender's SLIM name (from `MulticastItem.context.source`), per the spec's message-attribution requirement. Additive — not part of the point-to-point `Transport` trait, whose methods assume exactly one response per call.
- `register_collaborate` — registers a **listen-only** `Collaborate` participant on a `slim_rpc::Server`: invokes a callback for every inbound `Message`, sends nothing of its own (empty reply stream), matching the spec's non-participation mode. Independent of `RequestHandler`, since `Collaborate` is a broadcast channel rather than request/response.

**Not yet implemented** (spec is still draft): Agent Card extension declaration, formal EOS/non-participation signaling beyond the empty reply stream, and server-side per-message sender attribution (`Context`/`DecodedStream` don't expose it with the current API — attribution is only available client-side via `MulticastItem`, which is what `collaborate()` uses).

Bumps `a2a-slimrpc` to **0.2.3**; adds `serde_json`; promotes `tokio` to a real dependency (the SLIM stack already requires a tokio runtime transitively).

## Test

`slimrpc_collaborate_broadcasts_to_every_other_member`: 3 members each independently broadcast one intro `Message` to the other two via `collaborate()`, and every other member's `register_collaborate` listener receives it — proving many-to-many delivery end to end. All existing tests + clippy + fmt pass.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -124,7 +124,7 @@ dependencies = [
 
 [[package]]
 name = "a2a-slimrpc"
-version = "0.2.2"
+version = "0.2.3"
 dependencies = [
  "a2a-client-lf",
  "a2a-lf",
@@ -140,6 +140,7 @@ dependencies = [
  "futures",
  "prost",
  "prost-types",
+ "serde_json",
  "tokio",
 ]
 
--- a/a2a-slimrpc/Cargo.toml
+++ b/a2a-slimrpc/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "a2a-slimrpc"
-version = "0.2.2"
+version = "0.2.3"
 description = "A2A v1 SLIMRPC protocol binding for client and server"
 readme = "README.md"
 edition.workspace = true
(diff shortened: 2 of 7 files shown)


msardara (approved):
LGTM ✅

PR #113, opened 2026-07-23
test: use process restult scripts from itk repository

JakubWorek:
# Description 

Uses process results script provided from itk repository instead of custom one - as part of ITK consolidation


muscariello (approved):
LGTM

PR #114, opened 2026-07-26
fix(a2a-grpc): preserve precise A2A error codes

### LINKED ISSUE #103: gRPC client ignores google.rpc.Status/ErrorInfo, collapsing all FAILED_PRECONDITION errors to -32002 (TaskNotCancelable)
The `a2a-grpc` client maps incoming gRPC errors to A2A errors using **only** the
`tonic::Status` code. It never decodes the `google.rpc.Status` / `ErrorInfo` carried
in the `grpc-status-details-bin` trailer. Because A2A spec §5.4 maps three distinct
errors to the same gRPC status `FAILED_PRECONDITION`, the client cannot tell them
apart and unconditionally coerces all of them to `-32002` (`TaskNotCancelable`):

| A2A error                          | JSON-RPC | gRPC status         |
|------------------------------------|----------|---------------------|
| TaskNotCancelableError             | -32002   | FAILED_PRECONDITION |
| PushNotificationNotSupportedError  | -32003   | FAILED_PRECONDITION |
| UnsupportedOperationError          | -32004   | FAILED_PRECONDITION |

The specific error is meant to be recovered from the `ErrorInfo` detail
(`reason` = UPPER_SNAKE_CASE name, `domain` = `a2a-protocol.org`, AIP-193 rich error
model). The Go, Python, and .NET clients all read it; the Rust gRPC client does not.

### Location

`a2a-grpc/src/errors.rs`, `status_to_a2a_error` (lines 29–40):

```rust
pub fn status_to_a2a_error(status: &tonic::Status) -> A2AError {
    let code = match status.code() {          // only the gRPC code is inspected
        tonic::Code::NotFound => error_code::TASK_NOT_FOUND,
        tonic::Code::FailedPrecondition => error_code::TASK_NOT_CANCELABLE,  // fixed -32002
        ...
    };
    A2AError::new(code, status.message())      // status.details() never read
}
```
Every gRPC client method routes errors through this function
(a2a-grpc/src/client.rs, .map_err(|s| status_to_a2a_error(&s))).
status.details() is never called anywhere in the crate.

Steps to reproduce
Call CreateTaskPushNotificationConfig (gRPC) against a server that does not support
push notifications. The server returns FAILED_PRECONDITION with a correct
ErrorInfo (reason=PUSH_NOTIFICATION_NOT_SUPPORTED, domain=a2a-protocol.org).

Expected: A2A code -32003 (PushNotificationNotSupported)
Actual: A2A code -32002 (TaskNotCancelable)
Observed in the agntcy-a2a interop matrix (Rust client → .NET server, gRPC). Go and
Python clients recover -32003 correctly from the same response.

Note
The SDK already has the logic to fix this — a2a-client/src/lib.rs
a2a_error_from_details() reads ErrorInfo, checks domain == "a2a-protocol.org",
and maps reason via reason_to_error_code(). It is wired into the JSON-RPC and REST
clients (jsonrpc.rs, rest.rs) but not the gRPC client. The gRPC path should
decode status.details() as google.rpc.Status, extract the ErrorInfo, and reuse
reason_to_error_code() — mirroring the other transports.

jstar0:
## Summary

The gRPC boundary currently maps errors using only `tonic::Code`. Six distinct
A2A errors share `FAILED_PRECONDITION`, so clients decode all of them as
`TaskNotCancelable`. The server conversion also omits the
`google.rpc.ErrorInfo` detail that other SDKs need to recover the precise A2A
reason.

## Changes

- Emit an AIP-193 `ErrorInfo` with the existing A2A reason and
  `a2a-protocol.org` domain from the server conversion.
- Decode only known reasons from that protocol domain on the client conversion.
- Preserve string metadata from existing `ErrorInfo` details and field
  violations from existing `BadRequest` details when constructing the gRPC
  status.
- Keep the current gRPC-code fallback for peers that omit details or send
  malformed, unknown, or foreign-domain details.
- Cover all six `FAILED_PRECONDITION` mappings in both directions, including an
  independently constructed peer status, and cover all four
  `INVALID_ARGUMENT` reasons plus structured detail propagation.

This uses `tonic-types`, the official rich error model companion for the
workspace's existing Tonic 0.14 dependency.

## Compatibility

The gRPC status codes and messages are unchanged. Existing peers without rich
details continue to use the previous fallback behavior. Malformed detail
entries are ignored rather than changing status conversion.

## Verification

```bash
cargo test -p a2a-grpc errors
cargo hack check --workspace --all-targets --each-feature
cargo fmt --all -- --check
cargo hack clippy --workspace --all-targets --each-feature -- -D warnings
cargo test --workspace
```

Fixes #103

### DIFF
--- a/a2a-grpc/src/errors.rs
+++ b/a2a-grpc/src/errors.rs
@@ -3,7 +3,7 @@
 use std::collections::HashMap;
 
 use a2a::{A2AError, error_reason, errordetails, reason_to_error_code};
-use tonic_types::{ErrorDetails, StatusExt};
+use tonic_types::{ErrorDetails, FieldViolation as GrpcFieldViolation, StatusExt};
 
 /// Convert an A2A error to a tonic gRPC status code.
 pub fn a2a_error_to_status(err: &A2AError) -> tonic::Status {
@@ -25,11 +25,56 @@ pub fn a2a_error_to_status(err: &A2AError) -> tonic::Status {
         error_code::INTERNAL_ERROR => tonic::Code::Internal,
         _ => tonic::Code::Unknown,
     };
-    let details = ErrorDetails::with_error_info(
+    let metadata: HashMap<String, String> = err
+        .details
+        .as_deref()
+        .and_then(|details| {
+            details
+                .iter()
+                .find(|detail| detail.type_url == errordetails::ERROR_INFO_TYPE)
+        })
+        .and_then(|detail| detail.value.get("metadata"))
+        .and_then(|metadata| metadata.as_object())
+        .map(|metadata| {
+            metadata
+                .iter()
+                .filter_map(|(key, value)| {
+                    value.as_str().map(|value| (key.clone(), value.to_owned()))
+                })
+                .collect()
+        })
+        .unwrap_or_default();
+    let mut details = ErrorDetails::with_error_info(
         error_reason(err.code),
         errordetails::PROTOCOL_DOMAIN,
-        HashMap::new(),
+        metadata,
     );
+
+    let mut field_violations = Vec::new();
+    if let Some(error_details) = &err.details {
+        for detail in error_details
+            .iter()
+            .filter(|detail| detail.type_url == errordetails::BAD_REQUEST_TYPE)
+        {
+            let Some(violations) = detail
+                .value
+                .get("fieldViolations")
+                .and_then(|value| value.as_array())
+            else {
+                continue;
+            };
+            field_violations.extend(violations.iter().filter_map(|violation| {
+                Some(GrpcFieldViolation::new(
+                    violation.get("field")?.as_str()?,
+                    violation.get("description")?.as_str()?,
+                ))
+            }));
+        }
+    }
+    if !field_violations.is_empty() {
+        details.set_bad_request(field_violations);
+    }
+
     tonic::Status::with_error_details(code, &err.message, details)
 }
 
@@ -75,6 +120,15 @@ mod tests {
             "EXTENSION_SUPPORT_REQUIRED",
         ),
         (error_code::VERSION_NOT_SUPPORTED, "VERSION_NOT_SUPPORTED"),
+    ];
+    const INVALID_ARGUMENT_ERRORS: [(i32, &str); 4] = [
+        (
+            error_code::CONTENT_TYPE_NOT_SUPPORTED,
+            "CONTENT_TYPE_NOT_SUPPORTED",
+        ),
+        (error_code::PARSE_ERROR, "PARSE_ERROR"),
+        (error_code::INVALID_REQUEST, "INVALID_REQUEST"),
+        (error_code::INVALID_PARAMS, "INVALID_PARAMS"),
     ];
 
     #[test]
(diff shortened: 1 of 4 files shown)


a2a-grpc/src/errors.rs
```diff
@@ -22,27 +25,57 @@ pub fn a2a_error_to_status(err: &A2AError) -> tonic::Status {
...
+        (
+            error_code::EXTENDED_CARD_NOT_CONFIGURED,
+            "EXTENDED_AGENT_CARD_NOT_CONFIGURED",
+        ),
+        (
+            error_code::EXTENSION_SUPPORT_REQUIRED,
+            "EXTENSION_SUPPORT_REQUIRED",
+        ),
+        (error_code::VERSION_NOT_SUPPORTED, "VERSION_NOT_SUPPORTED"),
+    ];
```
msardara, line 123: I think here we should also add a constant for the 4 invalid arguments error:

```rust
const INVALID_ARGUMENT_ERRORS: [(i32, &str); 4] = [
    (error_code::CONTENT_TYPE_NOT_SUPPORTED, "CONTENT_TYPE_NOT_SUPPORTED"),
    (error_code::PARSE_ERROR, "PARSE_ERROR"),
    (error_code::INVALID_REQUEST, "INVALID_REQUEST"),
    (error_code::INVALID_PARAMS, "INVALID_PARAMS"),
];
```

And then add some tests like the ones added for the FAILED_PRECONDITION

a2a-grpc/src/errors.rs
```diff
@@ -22,27 +25,57 @@ pub fn a2a_error_to_status(err: &A2AError) -> tonic::Status {
         error_code::INTERNAL_ERROR => tonic::Code::Internal,
         _ => tonic::Code::Unknown,
     };
-    tonic::Status::new(code, &err.message)
+    let details = ErrorDetails::with_error_info(
+        error_reason(err.code),
+        errordetails::PROTOCOL_DOMAIN,
+        HashMap::new(),
```
msardara: I think here we should not just insert an empty dictionary in the error details. We can forward some details - like field violations - as we do [with jsonrpc errors](https://github.com/a2aproject/a2a-rs/blob/main/a2a/src/errors.rs#L190):

```rust
let metadata: HashMap<String, String> = err
    .details
    .as_ref()
    .and_then(|ds| {
        ds.iter()
            .find(|d| d.type_url == errordetails::ERROR_INFO_TYPE)
            .and_then(|d| d.value.get("metadata"))
            .and_then(|m| m.as_object())
            .map(|m| {
                m.iter()
                    .filter_map(|(k, v)| v.as_str().map(|s| (k.clone(), s.to_owned())))
                    .collect()
            })
    })
    .unwrap_or_default();

let mut error_details = ErrorDetails::with_error_info(
      error_reason(err.code),
      errordetails::PROTOCOL_DOMAIN,
      metadata,
  );

  if let Some(ds) = &err.details {
      let violations: Vec<FieldViolation> = ds
          .iter()
          .filter(|d| d.type_url == errordetails::BAD_REQUEST_TYPE)
          .filter_map(|d| d.value.get("fieldViolations")?.as_array().map(|a| a.to_vec()))
          .flatten()
          .filter_map(|v| {
              Some(FieldViolation {
                  field: v.get("field")?.as_str()?.to_owned(),
                  description: v.get("description")?.as_str()?.to_owned(),
              })
          })
          .collect();
      if !violations.is_empty() {
          error_details.set_bad_request(violations);
      }
  }

  tonic::Status::with_error_details(code, &err.message, error_details)
```

Then we should also add some tests for this, around the error info metadata and the field violations being reported to the grpc status.

msardara (changes requested):
Thanks for the PR @jstar0! PR looks good, but I added some comments regarding some information we potentially miss when constructing the grpc status.

a2a-grpc/src/errors.rs
jstar0: Added `INVALID_ARGUMENT_ERRORS` for `CONTENT_TYPE_NOT_SUPPORTED`, `PARSE_ERROR`, `INVALID_REQUEST`, and `INVALID_PARAMS`, with a round-trip test that verifies the gRPC code, `ErrorInfo` reason/domain, and recovered A2A code. This is in 833c3d8; the focused error suite passes 12/12 and the full `just ci` run passes.

a2a-grpc/src/errors.rs
jstar0: Updated `a2a_error_to_status` to forward string metadata from an existing `ErrorInfo` and valid field/description pairs from existing `BadRequest` details into tonic `ErrorDetails`. The regression checks both metadata entries and two field violations; malformed entries remain ignored. This is in 833c3d8, and the full `just ci` run passes.

msardara (approved):
LGTM, thanks @jstar0!

PR #116, opened 2026-07-29
build(itk): commit Cargo.lock for reproducible agent builds

JakubWorek:
Un-ignores and commits itk/Cargo.lock, regenerated with Rust 1.85.0 to match the ITK container. 
Verified with cargo build --locked --release. 
Inert until [the harness passes --locked](https://github.com/a2aproject/a2a-itk/pull/27).


msardara (approved):
LGTM

PR #122, opened 2026-08-10
fix(server): validate list task inputs (historyLength, status, pageSize, pageToken)

ez-lbz:
### 1. Reject invalid `status` filter with 400 instead of silently returning all tasks 

**Problem:** `handle_list_tasks` parsed the `?status=` query parameter with `serde_json::from_value(...).ok()`, so an unparseable value (e.g. `?status=invalid`) was silently dropped and the filter was not applied — the client received every task as if no filter was set. Python/JS reject invalid status values.

**Fix (a2a-server/src/rest.rs):**
- `handle_list_tasks` now returns a 400 `INVALID_PARAMS` error (`"invalid status filter"`) when the `status` query parameter cannot be parsed as a `TaskState`.

### 2. Treat negative `historyLength` as empty history instead of wrapping to `usize::MAX` 

**Problem:** `let hl = *hl as usize;` casts a negative `i32` (e.g. `-1`) to `usize::MAX` on 64-bit platforms, silently disabling history truncation and returning the full conversation history (information disclosure, same root cause as Go's ).

**Fix (a2a-server/src/task_store/inmemory.rs):**
- Added a shared `truncate_history(task, history_length)` helper: negative values (and `0`) truncate history to an empty list; positive values keep the last `hl` messages.
- The task-list path now uses this helper.

### 3. Cap `pageSize` at 100 (default 50) to prevent unbounded responses 

**Problem:** Any positive `pageSize` was accepted, so a client could request `pageSize=999999999` and force the server to serialize and transmit the entire task store in one response (resource exhaustion). Python caps at 100, JS at 50.

**Fix (a2a-server/src/task_store/inmemory.rs, a2a-server/src/handler.rs):**
- `InMemoryTaskStore::list` caps the page size at 100 (`MAX_PAGE_SIZE`), keeping the default of 50.
- `DefaultRequestHandler::list_push_configs` applies the same cap (max 100, default 50) to keep push-config pagination consistent.

### 4. Invalid `pageToken` now returns an error instead of silently resetting to page 0 

**Problem:** `token.parse::<usize>().unwrap_or(0)` silently treated a stale or corrupted page token as offset 0, returning the first page and causing clients to re-process already-seen tasks. Go/Python/JS return errors for invalid pagination tokens.

**Fix (a2a-server/src/task_store/inmemory.rs, a2a-server/src/handler.rs):**
- `InMemoryTaskStore::list` returns `INVALID_PARAMS` (`"invalid page token"`) when the token is not a valid `usize`.
- `DefaultRequestHandler::list_push_configs` does the same for push-config pagination.
- Fixed a latent panic: `tasks[start..end]` would panic if `start` exceeded the total size. The start offset is now clamped to `tasks.len()` in both places.

## Testing

- `cargo test -p a2a-server-lf`: **133 passed, 0 failed** (+ 1 doc-test passed).
- New regression tests:
 - `task_store::inmemory::tests::test_list_negative_history_length_returns_empty_history`
 - `task_store::inmemory::tests::test_list_page_size_is_capped`
 - `task_store::inmemory::tests::test_list_invalid_page_token_errors`
 - `task_store::inmemory::tests::test_list_page_token_beyond_total_clamps`
 - `handler::tests::test_list_push_configs_invalid_page_token_errors`
 - `handler::tests::test_list_push_configs_page_size_is_capped`
 - `rest::tests::test_list_tasks_invalid_status_returns_bad_request`
- Behavior change: an invalid `pageToken` (previously silently treated as offset 0) and an invalid `status` filter (previously silently ignored) now return 400. Existing test `rest::tests::test_list_tasks_with_query_params` was updated to use a valid `pageToken=0` since it previously relied on the lenient `pageToken=tok` behavior.
- Note: `examples` / `a2a-grpc` / `a2a-slimrpc` packages cannot be compiled in this environment (missing NASM for `aws-lc-sys` and missing `protoc`) — a pre-existing environment limitation, unrelated to this change.

### DIFF
--- a/a2a-server/src/handler.rs
+++ b/a2a-server/src/handler.rs
@@ -1385,7 +1385,9 @@ mod tests {
         install_crypto_provider();
         let handler = make_handler_with_push_configs();
         let params = ServiceParams::new();
-        for i in 0..120 {
+        // Keep the count below the per-task push config cap (50) so this test
+        // stays valid once push config limits are enforced.
+        for i in 0..40 {
             handler
                 .create_push_config(
                     &params,
@@ -1402,6 +1404,8 @@ mod tests {
                 .unwrap();
         }
 
+        // Requesting a huge page size must not error and returns everything
+        // available (the hard page-size cap of 100 is enforced in the store).
         let resp = handler
             .list_push_configs(
                 &params,
@@ -1414,7 +1418,7 @@ mod tests {
             )
             .await
             .unwrap();
-        assert_eq!(resp.configs.len(), 100);
+        assert_eq!(resp.configs.len(), 40);
     }
 
     #[tokio::test]
--- a/a2a-server/src/handler.rs
+++ b/a2a-server/src/handler.rs
@@ -750,16 +750,19 @@ impl RequestHandler for DefaultRequestHandler {
         let mut configs = self.push_config_store()?.list(&req.task_id).await?;
         configs.sort_by(|left, right| left.id.cmp(&right.id));
 
+        // Cap the page size (max 100, default 50) like the task list path.
         let page_size = match req.page_size {
-            Some(size) if size > 0 => size as usize,
+            Some(size) if size > 0 => (size as usize).min(100),
             _ => 50,
         };
-        let start = req
-            .page_token
-            .as_deref()
-            .and_then(|token| token.parse::<usize>().ok())
-            .unwrap_or(0)
-            .min(configs.len());
+        let start = if let Some(ref token) = req.page_token {
+            token
+                .parse::<usize>()
+                .map_err(|_| A2AError::invalid_params("invalid page token"))?
+        } else {
+            0
+        };
+        let start = start.min(configs.len());
         let end = (start + page_size).min(configs.len());
         let next_page_token = (end < configs.len()).then(|| end.to_string());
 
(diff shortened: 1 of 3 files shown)


muscariello:
Thanks @ez-lbz — all three validations here are real and none of them is on `main`. Plan is to merge #196 first and then rebase this, for a reason worth spelling out, plus a note on the conflict.

## The conflict with #140 is trivial

`a2a-server: clamp the page token offset when listing tasks` (#140) merged in August and fixes the same out-of-bounds slice in `InMemoryTaskStore::list`. I resolved it locally to check: `main`'s version uses `start.saturating_add(page_size)` where this branch uses `start + page_size`, so `main`'s is the one to keep. The remaining hunks are the two near-duplicate out-of-range-page-token tests, and both sets are worth keeping. Resolved that way the workspace is green — 26 suites, 0 failures, `fmt` and `cargo hack clippy --all-targets --each-feature -- -D warnings` clean.

## Why #196 goes first

Both PRs add a history-truncation helper, under different names:

| | Helper | Where | Called from |
| --- | --- | --- | --- |
| this PR | `truncate_history` | `task_store/inmemory.rs` | `list` only |
| #196 | `apply_history_length` | `task_store/mod.rs` | `get_task` **and** `list_tasks` |

git will not flag that as a conflict — the names differ and they live in different modules — so merging both leaves two functions doing the same job. #196's placement is the one to keep: #195 asks for "one shared `historyLength` normalization used by `get_task` and `list_tasks`", precisely so the get and list paths cannot drift apart again, which is the drift that produced the original bug.

So: #196 lands the shared helper, then this rebases on top and drops `truncate_history`, keeping the three validations that are unique to it.

## What remains unique here, and is worth having

- **An unparseable `status` filter is silently dropped.** `?status=invalid` currently returns *every* task as though no filter had been set. This is the one I'd call more than a papercut: a caller narrowing a list and receiving the unnarrowed set has no signal that it happened. Python and JS both reject it.
- **`pageSize` is unbounded** — now capped at 100.
- **An unparseable `pageToken` is treated as offset 0** — now `INVALID_PARAMS`.

Two of those change observable behaviour and should be called out in the release notes: a caller passing a garbage page token used to receive page 1 and will now receive an error, and one asking for `pageSize=500` will receive 100 plus a continuation token. Both are the correct behaviour and neither fails silently, which is the point.

## What happens next

No action needed from you unless you would rather drive it. Once #196 merges, a maintainer can rebase this branch (`maintainerCanModify` is enabled) — dropping `truncate_history` and taking `main`'s clamp — or you are very welcome to do it yourself if you prefer to keep the branch in your hands.

One question while it waits: rejecting a non-numeric `pageToken` fixes the silent-offset-0 behaviour, but it also pins the token format as a decimal offset, so a later move to an opaque cursor becomes a breaking change for anyone who started validating against it. Happy either way — worth a moment's thought before it is locked in.

ez-lbz:
Rebased onto current `main` (post-#196): dropped `truncate_history` in favour of the shared `apply_history_length` helper and kept `main`'s `saturating_add` clamp, so what remains here is exactly the three validations unique to this PR — strict `status` filter parsing at the REST boundary, the `pageSize` cap, and `INVALID_PARAMS` for unparseable `pageToken` — plus the corresponding regression tests. `cargo fmt`/`clippy --all-targets`/`cargo test` green locally (itk crate excluded — needs the ITK proto environment).

ez-lbz:
Heads-up on the DCO check (also applies to #124, #126, #127): it is red on all four of my PRs, but not for missing sign-offs — every commit on these branches now carries `Signed-off-by` (I force-pushed signed versions). The failure comes from the range the check walks: `BASE_SHA: ${{ github.event.pull_request.base.sha }}` is a snapshot from when the PR was created (these date to Aug 10) and is not refreshed by `synchronize` events, so the range sweeps in every `main` commit merged since — including unsigned upstream commits from the 9/13 batch (`425ec753` #151, `d156eeb0` #153, `baa86e21` #159, `4bbc7786` #160, `3740e3dd` #206, `572514a8` #163). The per-commit output confirms it: all of my commits report `ok`; the FAILs are upstream's own.

If you want the check robust for long-lived PRs, deriving the base in-workflow avoids the stale snapshot, e.g.:

```yaml
- name: Verify Developer Certificate of Origin
  run: |
    base=$(git merge-base origin/main \$HEAD_SHA\)
    commits=$(git rev-list --no-merges "$base".."$HEAD_SHA")
    ...
```

Happy to send that as a PR to `dco.yml` if useful. Alternatively a retarget of these PRs (base off `main` and back) should refresh the event's base snapshot.

muscariello (approved):
Approving. Verified locally at `5905533` on top of current `main`:

- `cargo fmt --all -- --check` clean
- `cargo test -p a2a-server-lf` — 155 passed, 0 failed
- `cargo clippy -p a2a-server-lf --all-targets -- -D warnings` clean
- Net diff reviewed: `truncate_history` is gone in favour of the shared
  `apply_history_length` from #196, and `main`'s `saturating_add` clamp from
  #140 is preserved. The three validations unique to this PR are what remain.

On the earlier question about pinning the page-token format as a decimal
offset: leaving it as-is. Rejecting a non-numeric token is the right behaviour
today, and a later move to an opaque cursor is a deliberate protocol change
that would need its own migration regardless.

One thing not addressed here, and not a blocker since it is strictly better
than `main` (which caps nothing anywhere): Copilot's point that the `pageSize`
cap lives in `InMemoryTaskStore` rather than in `DefaultRequestHandler`, so a
custom `TaskStore` still receives the caller's raw `page_size`. This PR already
clamps at the handler for `list_push_configs`, so the two list paths are
inconsistent. Filed as a follow-up rather than held here.

PR #123, opened 2026-08-10
fix(server): apply historyLength to get_task, make cancel idempotent, add progressive auth check

ez-lbz:
### 1. `get_task` now applies `historyLength` truncation 

**Problem:** `get_task` returned the full task history regardless of the `historyLength` field, while `list_tasks` truncated it. A client calling `getTask` with `historyLength=1` still received every prior message — an information-disclosure divergence from Python/JS/Go, which truncate in both `getTask` and `listTasks`.

**Fix (a2a-server/src/handler.rs):**
- `get_task` now truncates the returned task's history to the last `historyLength` messages, mirroring the `list_tasks` logic. `historyLength=0` returns an empty history; negative values are treated as an empty history ( semantics) instead of wrapping to a huge `usize`.

### 2. Canceling a terminal task is now idempotent 

**Problem:** `cancel_task` returned `TASK_NOT_CANCELABLE` when the task was already in a terminal state (COMPLETED, CANCELED, FAILED). TypeScript and Go treat cancel-on-terminal as idempotent and return the current terminal task, so a client retrying a cancel after a network timeout succeeded on one server and failed on another.

**Fix (a2a-server/src/handler.rs):**
- When the loaded task is already terminal, `cancel_task` returns the current terminal task (`Ok(task)`) instead of an error.
- Behavior change: existing test `handler::tests::test_cancel_task_already_completed` was updated to assert the idempotent success behavior.

### 3. Progressive authorization enforcement on handler methods 

**Problem:** Every handler method received `ServiceParams` (headers extracted by the middleware) but ignored them — no authorization check on `get_task`, `list_tasks`, `cancel_task`, or any other method. Any caller could read/cancel any task or list all tasks (CWE-862).

**Fix (a2a-server/src/handler.rs):**
- Added a progressive `authorize` method on `DefaultRequestHandler`. It is a **no-op by default**, so default deployments behave exactly as before (existing middleware / transport tests are unaffected).
- Added a builder `DefaultRequestHandler::require_authentication()` that opts in to enforcement: when enabled, every task-related method (`send_message`, `send_streaming_message`, `get_task`, `list_tasks`, `cancel_task`, `subscribe_to_task`, and all push-config methods) rejects requests that do not carry an `Authorization` header (surfaced through `ServiceParams`) with `INVALID_REQUEST: "authentication required"`.
- The `authorize` signature takes `(params, task_id)` — `task_id` is reserved for future owner/tenant-based checks once tasks carry ownership metadata; the plan's skipped (no owner concept) will be covered by this extension point.

## Testing

- `cargo test -p a2a-server-lf`: **130 passed, 0 failed** (+ 1 doc-test passed).
- New regression tests:
 - `handler::tests::test_get_task_applies_history_length`
 - `handler::tests::test_get_task_negative_history_length_returns_empty_history`
 - `handler::tests::test_require_authentication_rejects_unauthenticated_requests`
 - `handler::tests::test_require_authentication_allows_authenticated_requests`
- Updated existing test: `handler::tests::test_cancel_task_already_completed` (cancel-on-terminal now returns the terminal task instead of an error).
- Behavior change: only for handlers built with `require_authentication()`; default behavior is unchanged. `transports_e2e` uses a custom `RequestHandler` (not `DefaultRequestHandler`), so it is unaffected.

### DIFF
--- a/a2a-server/src/handler.rs
+++ b/a2a-server/src/handler.rs
@@ -424,6 +424,7 @@ pub struct DefaultRequestHandler {
     push_config_store: Option<Arc<dyn crate::PushConfigStore>>,
     push_sender: Option<Arc<crate::HttpPushSender>>,
     capabilities: AgentCapabilities,
+    require_authentication: bool,
 }
 
 impl DefaultRequestHandler {
@@ -435,6 +436,7 @@ impl DefaultRequestHandler {
             push_config_store: None,
             push_sender: None,
             capabilities: AgentCapabilities::default(),
+            require_authentication: false,
         }
     }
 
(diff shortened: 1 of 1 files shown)


muscariello:
Thanks @ez-lbz — closing this in favour of #196, which carries all three fixes forward. Tracked in #195.

Both of the behaviour fixes here were right and are kept:

- **`historyLength` on `get_task`.** `list_tasks` truncated and `get_task` did not, so a client asking for `historyLength=1` still received the whole conversation. #196 keeps that fix and shares one normalization helper between the get and list paths, so the two cannot drift apart again.
- **Cancel on a terminal task.** #196 narrows it to an already-`CANCELED` task rather than any terminal state, which is what a2a-go does (`internal/taskexec/distributed_manager.go:164` returns the task for `CANCELED`; `:168` still returns `ErrTaskNotCancelable` for `COMPLETED`/`FAILED`/`REJECTED`, and `promise.go:70` resolves the same way). a2a-cli's specification also sanctions the narrower reading — §10.4: "The operation is idempotent and MAY return a not-cancelable error if the task has already reached a terminal state."

The authorization piece went a different way, and it is worth saying why plainly rather than leaving it implied: `require_authentication()` accepted any non-empty `Authorization` value, so `Authorization: arbitrary` passed. That is a header-presence gate rather than authentication, and CWE-862 is not closed by it. #196 instead adds a `RequestAuthorizer` callback with a no-op default, so an application that already validates a token, DID, or session supplies that result — and the handler never invents a principal by observing that a header exists. The lowercase-only `"authorization"` lookup would also have rejected valid clients on SlimRPC, which copies metadata keys as-is.

The instinct was the right one — the methods genuinely had no authorization seat, and that is now fixed. Thanks for surfacing it.

ez-lbz:
Thanks for the detailed writeup — especially the CWE-862 point on the header-presence gate and the SlimRPC metadata-key caveat; the `RequestAuthorizer` seat is clearly the right shape. Good to see all three fixes landed via #196.

PR #124, opened 2026-08-10
fix(server): harden push notifications (URL validation, CRLF credentials, per-task config cap)

ez-lbz:
### 1. SSRF protection for push notification URLs 

**Problem:** `send_push` POSTed to whatever URL a client supplied in `TaskPushNotificationConfig` with no scheme/host/IP validation. A malicious client could point the server at `http://127.0.0.1`, `http://169.254.169.254` (cloud metadata), or any internal service (CWE-918). Go fixed this with an `ssrfGuardedTransport` + `isBlockedIP`; Python and JS fixed it via PR. Rust had no protection.

**Fix (a2a-server/src/push/sender.rs):**
- Added `validate_push_url(url)` invoked at the start of `send_push`: the scheme must be `http`/`https`; well-known loopback/metadata hosts (`127.0.0.1`, `localhost`, `::1`, `169.254.169.254`, `metadata.google.internal`, `metadata.azure.com`, `metadata.goog`, `0.0.0.0`) are blocked; IP literals in loopback, RFC 1918 private, link-local, IPv6 ULA (`fc00::/7`), unspecified, or multicast ranges are blocked.
- Added `HttpPushSenderConfig::validate_urls` (default `true` — secure by default). Disable only for trusted local webhooks; the flag is exposed for test environments.

### 2. Reject CR/LF in push credentials and tokens 

**Problem:** `Authorization` / `A2A-Notification-Token` headers were built by string concatenation, so credentials or tokens containing CR/LF could inject additional headers into the outgoing push request (header injection). `reqwest` rejects such values, but the error was opaque.

**Fix (a2a-server/src/push/sender.rs):**
- `send_push` now rejects credentials or notification tokens containing `\r` or `\n` with an explicit `INVALID_PARAMS` error before building the request.

### 3. Per-task push notification config limit 

**Problem:** `InMemoryPushConfigStore::save` accepted an unlimited number of configs per task; each config consumes memory and triggers an outbound HTTP request on every event, so a client could register unbounded configs (resource exhaustion).

**Fix (a2a-server/src/push/store.rs):**
- `InMemoryPushConfigStore::save` now enforces a cap of 50 configs per task (`MAX_PUSH_CONFIGS_PER_TASK`); exceeding it returns `INVALID_PARAMS`. Replacing an existing config by ID is still allowed at the cap.
- `HttpPushSenderConfig` is now exported from `a2a_server` (`lib.rs`, `push/mod.rs`) so consumers can construct senders with explicit configuration.

### 4. Test adaptation for the new URL validation

**Problem:** Existing tests deliver pushes to a webhook on `127.0.0.1`, which the new SSRF guard rejects.

**Fix (a2a-server/src/handler.rs tests, examples/tests/transports_e2e.rs):**
- `handler::tests::make_push_delivery_handler` and the e2e `spawn_push_http_server` now build the sender with `validate_urls: false`, explicitly opting out of URL validation for local test webhooks. The default `HttpPushSender::new(None)` keeps validation enabled.
- The connection-refused and auth-header tests in `push/sender.rs` now use `validate_urls: false` so they still exercise the HTTP path; the validation logic itself is covered by new dedicated tests.

## Testing

- `cargo test -p a2a-server-lf`: **135 passed, 0 failed** (+ 1 doc-test passed).
- New regression tests:
 - `push::sender::tests::test_validate_push_url_rejects_non_http_schemes`
 - `push::sender::tests::test_validate_push_url_rejects_loopback`
 - `push::sender::tests::test_validate_push_url_rejects_private_and_link_local`
 - `push::sender::tests::test_validate_push_url_accepts_public_urls`
 - `push::sender::tests::test_send_push_rejects_blocked_url_even_without_fail_on_error`
 - `push::sender::tests::test_send_push_rejects_crlf_credentials`
 - `push::sender::tests::test_send_push_rejects_crlf_token`
 - `push::store::tests::test_save_caps_configs_per_task`
 - `push::store::tests::test_save_can_replace_existing_config_at_cap`
- Behavior change: by default (`HttpPushSender::new(None)` / `HttpPushSenderConfig::default()`) push URLs targeting loopback/private/link-local/metadata addresses are now rejected with `INVALID_PARAMS`, and configs beyond 50 per task are rejected. Existing tests that pushed to `127.0.0.1` webhooks were adapted as described above.

### DIFF
--- a/a2a-server/src/push/sender.rs
+++ b/a2a-server/src/push/sender.rs
@@ -9,6 +9,7 @@ use std::time::Duration;
 pub struct HttpPushSender {
     client: reqwest::Client,
     fail_on_error: bool,
+    validate_urls: bool,
 }
 
 /// Configuration for [`HttpPushSender`].
@@ -17,6 +18,11 @@ pub struct HttpPushSenderConfig {
     pub timeout: Duration,
     /// If true, push sending errors abort execution.
     pub fail_on_error: bool,
+    /// If true (default), reject push URLs that target loopback, private,
+    /// link-local, multicast, unspecified, or cloud-metadata addresses, and
+    /// URLs whose scheme is not http/https. Disable only when pushing to
+    /// local webhooks in trusted test environments.
+    pub validate_urls: bool,
 }
 
 impl Default for HttpPushSenderConfig {
@@ -24,6 +30,7 @@ impl Default for HttpPushSenderConfig {
         HttpPushSenderConfig {
             timeout: Duration::from_secs(30),
             fail_on_error: false,
+            validate_urls: true,
         }
     }
 }
@@ -38,6 +45,7 @@ impl HttpPushSender {
         HttpPushSender {
             client,
             fail_on_error: config.fail_on_error,
+            validate_urls: config.validate_urls,
         }
     }
 
@@ -47,6 +55,29 @@ impl HttpPushSender {
         config: &TaskPushNotificationConfig,
         event: StreamResponse,
     ) -> Result<(), A2AError> {
+        if self.validate_urls {
+            validate_push_url(&config.url)?;
+        }
+
+        // Reject credentials containing CR/LF so they cannot inject
+        // additional headers into the outgoing request (BUG-34).
+        if let Some(ref token) = config.token {
+            if token.contains('\r') || token.contains('\n') {
+                return Err(A2AError::invalid_params(
+                    "push notification token must not contain CR/LF",
+                ));
+            }
+        }
+        if let Some(ref auth) = config.authentication {
+            if let Some(ref creds) = auth.credentials {
+                if creds.contains('\r') || creds.contains('\n') {
+                    return Err(A2AError::invalid_params(
+                        "push credentials must not contain CR/LF",
+                    ));
+                }
+            }
+        }
+
         let body = match serde_json::to_vec(&event) {
             Ok(b) => b,
             Err(e) => return self.handle_error(format!("failed to serialize event: {e}")),
@@ -98,6 +129,69 @@ impl HttpPushSender {
     }
 }
 
+/// Validate a push notification URL against SSRF targets (BUG-54).
+///
+/// The URL scheme must be http/https and the host must not resolve to a
+/// loopback, private, link-local, multicast, or unspecified address, nor to a
+/// well-known cloud metadata endpoint.
+fn validate_push_url(url: &str) -> Result<(), A2AError> {
+    let parsed =
+        reqwest::Url::parse(url).map_err(|_| A2AError::invalid_params("invalid push notification URL"))?;
+
+    if parsed.scheme() != "http" && parsed.scheme() != "https" {
+        return Err(A2AError::invalid_params(
+            "push URL must be http or https",
+        ));
+    }
+
+    if let Some(host) = parsed.host_str() {
+        // Block well-known loopback, metadata, and unspecified hosts.
+        let blocked = [
+            "127.0.0.1",
+            "localhost",
+            "::1",
+            "[::1]",
+            "169.254.169.254",
+            "metadata.google.internal",
+            "metadata.azure.com",
+            "metadata.goog",
+            "0.0.0.0",
+        ];
+        if blocked.contains(&host) {
+            return Err(A2AError::invalid_params("push URL targets blocked host"));
+        }
+        // Block IP literals in restricted ranges (loopback, RFC 1918
+        // private, link-local, ULA, unspecified, multicast).
+        if let Ok(ip) = host.parse::<std::net::IpAddr>() {
+            let blocked = match ip {
+                std::net::IpAddr::V4(v4) => {
+                    v4.is_loopback()
+                        || v4.is_private()
+                        || v4.is_link_local()
+                        || v4.is_unspecified()
+                        || v4.is_multicast()
+                }
+                std::net::IpAddr::V6(v6) => {
+                    // fc00::/7 unique local addresses (IPv6 "private").
+                    let is_ula = (v6.segments()[0] & 0xfe00) == 0xfc00;
+                    v6.is_loopback()
+                        || is_ula
+                        || v6.is_unicast_link_local()
+                        || v6.is_unspecified()
+                        || v6.is_multicast()
+                }
+            };
+            if blocked {
+                return Err(A2AError::invalid_params(
+                    "push URL targets private/loopback/link-local address",
+                ));
+            }
+        }
+    }
+
+    Ok(())
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
@@ -109,6 +203,7 @@ mod tests {
         let config = HttpPushSenderConfig::default();
         assert_eq!(config.timeout, Duration::from_secs(30));
         assert!(!config.fail_on_error);
+        assert!(config.validate_urls);
     }
 
     #[test]
@@ -116,6 +211,7 @@ mod tests {
         install_crypto_provider();
         let sender = HttpPushSender::new(None);
         assert!(!sender.fail_on_error);
+        assert!(sender.validate_urls);
     }
 
     #[test]
@@ -124,9 +220,11 @@ mod tests {
         let config = HttpPushSenderConfig {
             timeout: Duration::from_secs(10),
             fail_on_error: true,
+            validate_urls: false,
         };
         let sender = HttpPushSender::new(Some(config));
         assert!(sender.fail_on_error);
+        assert!(!sender.validate_urls);
     }
 
     #[test]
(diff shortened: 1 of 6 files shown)


muscariello:
Parts 2 and 3 look right to me — rejecting CR/LF in credentials and tokens before the request is built, and capping configs per task, are both straightforward and match what the other SDKs do.

Part 1, the SSRF guard, is the one I'd push back on: as written it is a **URL-string** check, and that is specifically the approach a2a-go rejected. From `a2asrv/push/sender.go:93`:

> `ssrfGuardedTransport` clones the default transport and validates the resolved IP of every connection **before the TCP handshake**. Because the check runs at dial time it also covers **DNS-rebinding** targets and **redirect hops** (each hop re-dials), which **a URL-string check alone cannot catch**.

Two bypasses follow from that, both of which reach the exact targets the PR sets out to block:

1. **A hostname that resolves to a blocked address.** `validate_push_url` blocks IP *literals* and a fixed list of well-known names, so `http://127.0.0.1` and `http://169.254.169.254` are caught. But a client-supplied hostname that resolves to `127.0.0.1` or to the metadata IP is neither an IP literal nor on the list, and passes. Registering such a name is trivial (the `*.localtest.me` class of wildcard DNS), and this is the standard DNS-rebinding shape.

2. **Redirects are not checked at all.** The URL is validated once, before the request. An allowed host answering `302 Location: http://169.254.169.254/...` is followed by `reqwest`'s default redirect policy straight to the blocked target, with no second validation.

a2a-go also does two things this PR does not, both worth carrying over:

- **Bounds the redirect chain** — `maxPushRedirects = 10` (`sender.go:34`, `:83`, `:147`).
- **Drops the notification token on a cross-host redirect** (`sender.go:152`). The reasoning in its comment is worth reproducing: Go strips `Authorization` on cross-host redirects but *not* custom headers, so `A2A-Notification-Token` would otherwise leak to whatever host the redirect names. `reqwest` behaves the same way here, so the Rust sender has the same leak — and it is reachable without any SSRF, just a webhook that redirects.

## Suggestion

Validate at connect time rather than on the string. In `reqwest` that means either a custom connector that inspects the resolved `SocketAddr` before connecting, or resolving the host explicitly, checking every returned address, and connecting to a pinned address. Either way the check then covers rebinding and every redirect hop, as Go's does. Adding `redirect::Policy::limited(...)` and stripping the token on a cross-host hop closes the remaining two.

If dial-time validation is more than you want to take on in this PR, it would still be a real improvement to land parts 2 and 3 plus the redirect limit and token-stripping, and say plainly in the docs and the changelog that URL validation blocks literal addresses only and is **not** a complete SSRF defence. What I would avoid is shipping it described as SSRF protection, because the next reader reasonably stops looking — and `validate_urls: true` being the default makes it read as solved.

For context on why this matters beyond the abstract: `HttpPushSenderConfig::validate_urls` defaults to `true`, and the tests opt out with `validate_urls: false` to reach a `127.0.0.1` webhook. So the guard is off in exactly the paths that exercise delivery, and the code path that runs in production is the one with the least test coverage. Dedicated tests for the validator itself help, but they cannot catch a bypass that lives in the resolution step the validator never sees.

muscariello:
+ make sure codecov is happy 
thanks @ez-lbz

ez-lbz:
Codecov gap closed — the 12 uncovered lines were the IPv6 branch of `validate_push_url`, and writing tests for it exposed that the branch was dead code: `host_str()` serializes IPv6 hosts with brackets (`[fc00::1]`), which never parses as an `IpAddr`, so every IPv6 range restriction was silently skipped. Fixed by stripping the brackets before the literal parse, plus:

- IPv4-mapped IPv6 literals (`::ffff:127.0.0.1`) now inherit the IPv4 restrictions (mapped-literal bypass, flagged by Copilot review)
- new tests: invalid/empty-authority URLs, IPv6 ULA/link-local/multicast/unspecified rejection, IPv4-mapped rejection, IPv4 multicast rejection, global IPv6 acceptance

Patch coverage for `push/sender.rs` is now 100% of the validation paths. Not addressed here (needs a policy call): DNS rebinding and redirect-following — a public webhook hostname can still resolve to, or redirect to, a restricted address at request time; that wants per-hop re-validation or a no-redirect client policy, which I'm happy to take on in a follow-up if you want it in scope.

a2a-server/src/push/sender.rs
```diff
@@ -98,6 +129,85 @@ impl HttpPushSender {
...
+fn validate_push_url(url: &str) -> Result<(), A2AError> {
+    let parsed = reqwest::Url::parse(url)
+        .map_err(|_| A2AError::invalid_params("invalid push notification URL"))?;
+
+    if parsed.scheme() != "http" && parsed.scheme() != "https" {
+        return Err(A2AError::invalid_params("push URL must be http or https"));
+    }
+
+    if let Some(host) = parsed.host_str() {
+        // Block well-known loopback, metadata, and unspecified hosts.
```
muscariello, line 146: A single trailing dot bypasses this blocklist. `localhost.` is an equivalent fully-qualified form and resolves exactly as `localhost` does, so appending one character defeats the loopback and metadata entries.

```
block  http://127.0.0.1/x
block  http://LOCALHOST/x
ALLOW  http://localhost./x                 <--
ALLOW  http://metadata.google.internal./x  <--
block  http://127.0.0.1./x
block  http://169.254.169.254./x
```

Only the *hostname* entries are affected — IP literals keep working because `Url::parse` normalises a trailing dot away for those.

**Tracked as #223, which we are fixing ourselves — no action needed from you here.** This PR merged with the gap open deliberately: there is no push URL validation on `main` before it, so it is strictly better than the status quo and nothing regresses. The fix lands separately so it carries its own regression tests rather than restarting CI on this branch.

The suggested change, for reference:

```suggestion
        // A trailing dot is an equivalent fully-qualified form: "localhost."
        // resolves exactly as "localhost" does. Normalise it away, or every
        // hostname below is bypassed by appending one character.
        let host = host.strip_suffix('.').unwrap_or(host);
        // Block well-known loopback, metadata, and unspecified hosts.
```

muscariello (commented):
Reviewed the hardening in detail. The three fixes are all real gaps against Go/Python/JS, and the URL validation is more careful than most attempts at this — the IPv6 bracket-stripping and the IPv4-mapped (`::ffff:a.b.c.d`) case are exactly the two things people usually miss, and you got both. The numeric-form bypasses I went looking for are all closed by URL normalisation:

```
block  http://2130706433/x    (decimal 127.0.0.1)
block  http://0x7f.0.0.1/x    (hex octets)
block  http://127.1/x         (short form)
block  http://[::ffff:127.0.0.1]/x
```

**Merged.** Two follow-ups came out of this review, both filed and both ours to fix — nothing is being asked of you:

| Issue | What |
| --- | --- |
| #223 | A trailing dot (`localhost.`) bypasses the hostname entries in the blocklist. One-line fix, suggested inline above. |
| #224 | The guard inspects the URL string and never resolves the hostname, so CWE-918 is still open in the general case. |

## On the comparison with Go

The description cites Go's `ssrfGuardedTransport` + `isBlockedIP` as precedent, but those differ in kind from what is here. Go checks the **resolved IP at connect time**; this checks the **URL string at parse time**. So a hostname that resolves to a private address passes:

```
ALLOW  http://internal.corp.test/x
```

Internal DNS names and DNS rebinding go straight through. That is not a criticism of this PR — string validation stops the obvious cases and is strictly better than the nothing we had — but it is why #224 exists, and why the release notes should say what the check does and does not cover rather than implying parity.

## Behaviour change worth release-noting

`validate_urls` defaults to `true`, so any existing deployment pushing to a loopback or RFC 1918 webhook starts getting `INVALID_PARAMS`. That is the right default and the opt-out is there, but it will surprise someone on upgrade. Captured in the squash commit message.

## Minor

`// Maximum number of push notification configs allowed per task (BUG-42).` — `BUG-42` does not resolve to anything in this repository's tracker. Rolled into #223's scope rather than held.

## Note for the #122 follow-up

`MAX_PUSH_CONFIGS_PER_TASK = 50` makes the push-config page-size cap of 100 unreachable through `InMemoryPushConfigStore` — 50 < 100, so `.min(100)` in `list_push_configs` can never bind for that store. The cap still matters for a custom `PushConfigStore`; tracked in #222.

Thanks for this one — the three issues it closes are exactly the ones the other SDKs had to fix, and finding them independently in the Rust port is genuinely useful work.

muscariello (approved):
Approving. All three hardening measures are real gaps against Go/Python/JS, and
the implementation is careful where it counts — IPv6 bracket stripping and
IPv4-mapped (`::ffff:a.b.c.d`) handling are the two cases usually missed, and
both are right. Verified locally at `d364341`: `cargo test -p a2a-server-lf`
147 passed / 0 failed, `clippy --all-targets -D warnings` clean, `fmt` clean.

Merging with the trailing-dot bypass still open, deliberately: there is no push
URL validation on `main` at all today, so this is strictly better than the
status quo and nothing regresses. The one-line fix and its regression tests
follow immediately in a separate PR rather than restarting CI on this branch —
the suggestion stays on the record above for the reasoning.

The DNS-resolution gap is a follow-up, not a defect in this PR: string
validation cannot see where a hostname resolves, so a connect-time guard on the
resolved address is what actually closes CWE-918. Filing that separately.

PR #125, opened 2026-08-10
fix(server): add SSE event ids and keep-alive, enforce request body size limit

ez-lbz:
### 1. SSE events now carry monotonic IDs and a keep-alive 

**Problem:** Both SSE stream functions used `Event::default().data(...)` — no event `id` and no heartbeat. Proxies/load balancers can terminate idle SSE connections without heartbeats, and clients cannot implement resumable subscriptions without event IDs. Go and JS assign event IDs; Rust had neither.

**Fix (a2a-server/src/sse.rs):**
- `sse_from_stream` and `sse_jsonrpc_stream` now assign a monotonic, counter-based `id` to every event (`id: 1`, `id: 2`, ...), so clients can detect gaps and resume interrupted subscriptions.
- Both responses now attach `KeepAlive::new().interval(15s)` — axum emits an SSE comment frame (`:\n\n`) every 15 seconds to keep idle connections alive. (Note: `KeepAlive::text(":\n")` from the plan panics in axum 0.8 because comment text must not contain newlines; the built-in default comment frame is used instead.)

### 2. Explicit request body size limit of 10 MB 

**Problem:** No REST or JSON-RPC handler enforced a maximum request body size; a client could send a multi-GB JSON payload to exhaust server memory (same issue as across SDKs).

**Fix (a2a-server/src/rest.rs, a2a-server/src/jsonrpc.rs):**
- `rest_router` and `jsonrpc_router` now apply `DefaultBodyLimit::max(10 * 1024 * 1024)` (10 MB), so oversized request bodies are rejected with `413 Payload Too Large` instead of relying on the framework's implicit default.

## Testing

- `cargo test -p a2a-server-lf`: **128 passed, 0 failed** (+ 1 doc-test passed).
- New regression tests:
 - `sse::tests::test_sse_from_stream_ok` / `..._err` (updated): assert `id: 1` is present in the body.
 - `sse::tests::test_sse_jsonrpc_stream_ok` / `..._err` / `..._multiple_events` (updated): assert event IDs are present and monotonic (`id: 1`, `id: 2`, `id: 3`).
 - `rest::tests::test_send_message_body_over_limit_rejected`: an ~11 MB request body returns `413 PAYLOAD_TOO_LARGE`.
 - `jsonrpc::tests::test_send_message_body_over_limit_rejected`: same for the JSON-RPC binding.
- Behavior change: SSE responses now include `id:` fields and keep-alive comment frames; request bodies larger than 10 MB are rejected with 413. Both are additive hardening changes.

### DIFF
--- a/a2a-server/src/jsonrpc.rs
+++ b/a2a-server/src/jsonrpc.rs
@@ -6,7 +6,7 @@ use a2a::*;
 use a2a_pb::protojson_conv::{self, ProtoJsonPayload};
 use axum::{
     Json,
-    extract::State,
+    extract::{DefaultBodyLimit, State},
     http::{HeaderMap, StatusCode},
     response::IntoResponse,
 };
@@ -38,6 +38,9 @@ pub fn jsonrpc_router<H: RequestHandler>(handler: Arc<H>) -> axum::Router {
     let state = JsonRpcState { handler };
     axum::Router::new()
         .route("/", axum::routing::post(handle_jsonrpc::<H>))
+        // Cap the request body size at 10 MB instead of relying on the
+        // framework's implicit default (BUG-60).
+        .layer(DefaultBodyLimit::max(10 * 1024 * 1024))
         .with_state(state)
 }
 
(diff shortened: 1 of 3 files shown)


muscariello:
Thanks @ez-lbz — closing this in favour of #202, which carries all three fixes forward. Tracked in #201.

All three were real and all three are kept: SSE events had no `id`, there was no keep-alive, and neither router bounded the request body. a2a-go has all three (`internal/sse/sse.go:78` writes an `id:` per event, `:68` is `WriteKeepAlive`, `:39` caps at 10 MB), so this was a straight interop gap.

Two of your calls I kept over the alternatives, and it is worth recording why:

- **The monotonic counter, rather than a2a-go's per-event UUID.** Both satisfy `Last-Event-ID`, but only a counter makes a gap *detectable* — a client that receives 3 then 5 knows it missed one, which an opaque identifier cannot tell it. #202 keeps the counter and notes the deliberate divergence from Go.
- **`KeepAlive::new().interval(..)` rather than `KeepAlive::text(":\n")`.** Your note that the latter panics in axum 0.8 because comment text must not contain newlines saved a real trap.

#202 adds two properties neither version asserted, both now covered by tests: ids are scoped to a single response, so concurrent subscriptions do not share a numbering space; and error events are numbered too, since a client counting events must not see a gap merely because one of them reported a failure.

Worth knowing where this landed: it unblocks `a2acli`'s Tier 2 stream resumption (#186). `SPEC.md` §9.4 says to resume "from the last received event **where the server supports it**" — without event ids a2a-rs did not support it, so the CLI could never do better than re-subscribing from scratch against our own server. And the missing keep-alive is what was dropping those subscriptions to begin with. So this one had more reach than its diff suggests.

PR #126, opened 2026-08-10
fix(server): protect terminal task states and make cancel transition atomic

ez-lbz:
### 1. Terminal task states can no longer be blindly overwritten 

**Problem:** `InMemoryTaskStore::update` (and `handler::save_task`) wrote whatever state arrived, with no state-machine validation. A task in `COMPLETED` could be overwritten to `SUBMITTED` or `CANCELED` by a direct state assignment or a racing executor event.

**Fix (a2a-server/src/task_store/inmemory.rs):**
- `InMemoryTaskStore::update` now rejects any update that changes the state of a terminal task (`COMPLETED`, `FAILED`, `CANCELED`, `REJECTED`) with `INVALID_REQUEST`. Re-applying the *same* terminal state is still allowed so idempotent re-emission (e.g. cancel flows that re-emit the terminal event) keeps working; artifact/message-only updates on a terminal task remain possible, but the state itself is immutable.

### 2. Cancel check-then-act race made atomic 

**Problem:** `cancel_task` implemented cancel as check-then-act: it read the task, checked it was not terminal, then let the executor drive store writes. Two concurrent cancel requests could both pass the check; a cancel could also race with task completion, with no atomicity between the check and the state transition.

**Fix (a2a-server/src/task_store/store.rs, a2a-server/src/task_store/inmemory.rs, a2a-server/src/handler.rs):**
- Added `TaskStore::begin_cancel(task_id)` — atomically checks the task is non-terminal and transitions it to `CANCELED` under one write lock (returns `TASK_NOT_CANCELABLE` if already terminal, `TASK_NOT_FOUND` if missing). The trait provides a non-atomic default implementation (get + check + update) so existing custom stores keep compiling; `InMemoryTaskStore` overrides it with the lock-held version.
- `cancel_task` now: (1) returns the current task idempotently if it is already terminal; (2) performs the atomic `begin_cancel` transition; (3) if `begin_cancel` races with completion or another cancel (`TASK_NOT_CANCELABLE`), reloads and returns the current task idempotently; (4) drives the executor's cancel stream to completion as before.

## Testing

- `cargo test -p a2a-server-lf`: **132 passed, 0 failed** (+ 1 doc-test passed).
- New regression tests:
 - `task_store::inmemory::tests::test_update_terminal_state_rejected`
 - `task_store::inmemory::tests::test_update_terminal_same_state_allowed`
 - `task_store::inmemory::tests::test_begin_cancel_transitions_non_terminal`
 - `task_store::inmemory::tests::test_begin_cancel_terminal_rejected`
 - `task_store::inmemory::tests::test_begin_cancel_not_found`
 - `handler::tests::test_cancel_task_concurrent_both_succeed_idempotently` (two concurrent cancels both succeed and end in `CANCELED`)
- Updated existing test: `handler::tests::test_cancel_task_already_completed` now asserts cancel-on-terminal returns the current `COMPLETED` task (idempotent success) instead of an error — same behavior change as (the two PRs touch this test identically).
- Behavior change: direct updates that change a terminal task's state are now rejected (`INVALID_REQUEST`); concurrent cancels and cancel-vs-completion races now resolve idempotently instead of racing. Normal task lifecycle (Submitted → Working → Completed) is unaffected.

### DIFF
--- a/a2a-server/src/handler.rs
+++ b/a2a-server/src/handler.rs
@@ -651,11 +651,27 @@ impl RequestHandler for DefaultRequestHandler {
         params: &ServiceParams,
         req: CancelTaskRequest,
     ) -> Result<Task, A2AError> {
-        let task = self.load_task(&req.id).await?;
-
-        if task.status.state.is_terminal() {
-            return Err(A2AError::task_not_cancelable(&req.id));
+        // Idempotent: canceling a task that is already in a terminal state
+        // returns the current terminal task instead of an error (BUG-52).
+        if let Some(task) = self.task_store.get(&req.id).await? {
+            if task.status.state.is_terminal() {
+                return Ok(task);
+            }
+        } else {
+            return Err(A2AError::task_not_found(&req.id));
         }
+
+        // Atomically transition the task to CANCELED so two concurrent cancel
+        // requests cannot both pass the check-then-act window (BUG-44).
+        let task = match self.task_store.begin_cancel(&req.id).await {
+            Ok(task) => task,
+            Err(error) if error.code == error_code::TASK_NOT_CANCELABLE => {
+                // Raced with another cancel or with task completion: return
+                // the current task (idempotent).
+                return self.load_task(&req.id).await;
+            }
+            Err(error) => return Err(error),
+        };
 
         let active_execution = self.execution_manager.get(&req.id).await;
 
@@ -1205,6 +1221,43 @@ mod tests {
     }
 
     #[tokio::test]
+    async fn test_cancel_task_concurrent_both_succeed_idempotently() {
+        let handler = Arc::new(make_handler());
+        let params = ServiceParams::new();
+        let task = Task {
+            id: "t-race".into(),
+            context_id: "c-race".into(),
+            status: TaskStatus {
+                state: TaskState::Working,
+                message: None,
+                timestamp: None,
+            },
+            artifacts: None,
+            history: None,
+            metadata: None,
+        };
+        handler.task_store.create(task).await.unwrap();
+
+        let handlers = vec![Arc::clone(&handler), Arc::clone(&handler)];
+        let mut results = Vec::new();
+        for handler in handlers {
+            let params = params.clone();
+            let req = CancelTaskRequest {
+                id: "t-race".into(),
+                metadata: None,
+                tenant: None,
+            };
+            results.push(tokio::spawn(async move {
+                handler.cancel_task(&params, req).await
+            }));
+        }
+        for result in results {
+            let task = result.await.unwrap().unwrap();
+            assert_eq!(task.status.state, TaskState::Canceled);
+        }
+    }
+
+    #[tokio::test]
     async fn test_cancel_task_already_completed() {
         let handler = make_handler();
         let params = ServiceParams::new();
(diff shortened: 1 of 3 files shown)


muscariello:
Checked this against current `main`: it rebases cleanly, the workspace is green (26 suites, 0 failures) and `cargo hack clippy --all-targets --each-feature -- -D warnings` is clean. Both fixes look well judged to me — terminal-state immutability that still permits re-applying the *same* state, so idempotent re-emission from cancel flows keeps working, and `begin_cancel` holding the write lock across the check and the transition, which is the right shape for a check-then-act race.

One thing to reconcile before this merges, because it will not show up as a git conflict.

This PR also makes cancel idempotent for **any** terminal state:

```rust
if task.status.state.is_terminal() {
    return Ok(task);
}
```

#196 deliberately narrowed that to already-`CANCELED` only, because the broader form diverges from every other SDK:

- **a2a-go** returns the task for `CANCELED` (`internal/taskexec/distributed_manager.go:164`) and still returns `ErrTaskNotCancelable` for `COMPLETED`/`FAILED`/`REJECTED` (`:168`); the local manager resolves the same way (`promise.go:70`).
- **a2a-python** is stricter still — it errors on every terminal state, `CANCELED` included (`default_request_handler.py:136`).
- **a2a-cli's specification** sanctions the narrow reading explicitly, §10.4: *"The operation is idempotent and MAY return a not-cancelable error if the task has already reached a terminal state."* "Idempotent" there means a retried cancel creates no new effects — which is exactly Go's `CANCELED → return task` — not that any terminal state must succeed.

So whichever of the two lands second needs that block reconciled, otherwise merging this after #196 silently restores the divergence: cancelling a `COMPLETED` task would succeed here and return `TASK_NOT_CANCELABLE` from Go and Python.

Concretely, the terminal check in `cancel_task` wants to be:

```rust
if let Some(task) = self.task_store.get(&req.id).await? {
    if task.status.state == TaskState::Canceled {
        return Ok(task);
    }
    if task.status.state.is_terminal() {
        return Err(A2AError::task_not_cancelable(&req.id));
    }
} else {
    return Err(A2AError::task_not_found(&req.id));
}
```

`begin_cancel` already returns `TASK_NOT_CANCELABLE` for any terminal state, so it composes with that unchanged, and the race fallback below it — returning the current task when `begin_cancel` loses the race — stays correct either way.

The `test_cancel_task_concurrent_both_succeed_idempotently` test is a good one to keep; it would want a sibling asserting that a `COMPLETED` task still refuses.

ez-lbz:
Reconciled with #196 as suggested: `cancel_task` now takes the narrow reading — an already-`CANCELED` task returns the task (idempotent), any other terminal state returns `TASK_NOT_CANCELABLE` — and the `begin_cancel` race path applies the same semantics (another cancel winning returns the task; a completion racing in propagates the not-cancelable error). Merged current `main` (which carries #196's handler shape and authorize seat) and kept the `begin_cancel` TOCTOU serialization plus terminal-state immutability as this PR's unique parts. All cancel-path tests green locally, including `test_cancel_task_concurrent_both_succeed_idempotently` and the failed/rejected not-cancelable cases.

muscariello:
@ez-lbz can you check codecov ?

muscariello (approved):
LGTM

ez-lbz:
The 3 uncovered lines are the trait default for `begin_cancel` — the non-atomic get→check→update fallback, dead in-tree since every real store overrides it. Covering it would mean writing tests for code that shouldn't exist, and #252 removes the default outright, which closes the coverage gap along with the race itself. So I'd leave this one to #252 rather than add throwaway tests — happy to rebase #126 onto post-#252 `main` if the ordering needs it.

PR #128, opened 2026-08-11
fix: fallback to legacy REST subscribe path

muscariello:
Improve HTTP+JSON interop for SubscribeToTask by retrying /tasks/{id}/subscribe when /tasks/{id}:subscribe is not supported.


msardara (approved):
LGTM ✅

PR #132, opened 2026-08-24
chore(deps): move to the SLIM 2.3 crates

muscariello:
`slim_auth` 0.14.6 caps below 0.15, which `agntcy-slim-bindings` 2.3.0 and
`slim_rpc` 2.3.0 require. Anything depending on both ends up with two
`slim_auth` versions in the tree, and the bounds stop matching:

```
the trait bound `AuthProvider: slim_auth::traits::TokenProvider` is not satisfied
  --> a2a-slimrpc-0.2.4/src/client.rs:382
```

| crate | before | after |
|---|---|---|
| agntcy-slim-rpc | 2.0.0 | 2.3.0 |
| agntcy-slim-config | 0.14.5 | 0.16.0 |
| agntcy-slim-service | 0.12.6 | 0.12.10 |
| agntcy-slim-datapath | 0.18.2 | 0.18.6 |
| agntcy-slim-auth | 0.14.6 | 0.15.3 |

No source changes were needed.

- [x] 19 tests pass, clippy clean, workspace builds

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -147,13 +147,14 @@ dependencies = [
 
 [[package]]
 name = "agntcy-slim-auth"
-version = "0.14.6"
+version = "0.15.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "15858803a36af4ca6183658ff8156a0fb41d91f4fa93bf24fd63174ef5292f16"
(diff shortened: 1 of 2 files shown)


msardara (approved):
LGTM ✅

PR #133, opened 2026-08-24
ci: add Renovate for dependency updates

muscariello:
No dependency automation today — no Renovate, no Dependabot config.

Covers cargo across the workspace (10 manifests, 190 deps) and the 39
SHA-pinned actions (8 workflows, 53 deps). Weekly, with a dry run available via
dispatch. Uses the `PROJECT_APP_ID` / `PROJECT_APP_KEY` App the winget workflow
already uses.

The SLIM crates are grouped, because that lockstep is what #132 had to fix:
`slim_auth` 0.14 caps below 0.15 while `slim_rpc` 2.3 requires it, so a
single-crate bump puts two versions in the tree and breaks the trait bounds.

`Formula/a2acli.rb` is generated by the release automation, so the homebrew
manager is off — Renovate edits there would be overwritten.

The App needs the **workflows** permission to update `.github/workflows`;
without it the actions manager is skipped.

- [x] config validated against the latest Renovate; extraction dry-run clean


msardara (approved):
LGTM ✅

PR #146, opened 2026-08-27
fix(a2a-client): surface JSON-RPC errors on streaming calls

### LINKED ISSUE #134: [Bug]: a JSON-RPC error on a streaming call arrives as an empty stream (a2acli stream/subscribe print nothing, exit 0)
### What happens

A streaming call made over the JSON-RPC binding returns an *empty stream* when the server answers with a JSON-RPC error. `a2acli stream` and `a2acli subscribe` then print nothing and exit 0, so the reason the call failed — which the server did send — never reaches the caller.

### Why

`a2a-client/src/jsonrpc.rs`'s `call_streaming` hands the response body straight to the SSE parser, without looking at the status or the content type:

```rust
let response = builder
    .json(&rpc_request)
    .send()
    .await
    .map_err(|e| A2AError::internal(format!("HTTP request failed: {e}")))?;

let stream = response.bytes_stream();
let event_stream = parse_sse_stream(stream);
Ok(event_stream)
```

A JSON-RPC error answering a streaming method is HTTP 200 with an `application/json` body — the transport succeeded, the call did not — so `parse_sse_stream` finds no `data:` frames and the stream ends having yielded nothing. The unary `call` on the same type does check `rpc_response.error` (`jsonrpc.rs:78`); only the streaming path skips it. `send_streaming_message` and `subscribe_to_task` both route through `call_streaming`, so both are affected.

### Reproducing

Any server that refuses a streaming call will do. Two we hit:

- `subscribe` on an already-terminal task — `helloworld-server` answers `-32001` and the CLI prints nothing, exit 0.
- `stream` against a server with no streaming backend (`-32004`, unsupported operation) — same silence.

Observed on 2026-08-21 against `7676ec9` (`a2acli` 0.1.5) while cross-testing an independent Rust implementation of the protocol. `a2a-client/src/jsonrpc.rs` is byte-identical at `78eafbf` (`a2a-cli` 0.1.8), so it still stands.

### Only the JSON-RPC binding, as far as we saw

The REST binding surfaced the same refusals correctly in those runs — an error arrives there with a non-200 status — so this looks specific to JSON-RPC, where an error is always HTTP 200.

### The fix that worked on our side

We had the identical bug and fixed it by diverting on the response's content type: anything that is not `text/event-stream` is decoded as a JSON-RPC envelope and its error mapped through the usual conversion, while a response with *no* content type is still read as a stream — consuming the body to inspect it would hang on a stream that has yet to say anything, so the content type is what makes the check safe.

msardara:
## Summary
- `call_streaming` fed the response body straight to the SSE parser regardless of content type. A JSON-RPC error answering a streaming call is HTTP 200 with a plain `application/json` body (JSON-RPC reports errors in the envelope, not via status), so the SSE parser found no `data:` frames and yielded an empty stream — the error was silently dropped.
- Now checks `Content-Type` before parsing: anything other than `text/event-stream` is decoded as a `JsonRpcResponse` and its error mapped through the existing conversion; a response with no content type is still treated as a stream (reading it to check would hang on a stream that hasn't sent anything yet).
- REST and gRPC bindings are unaffected — REST errors carry a non-2xx status, and gRPC refusals surface via the `grpc-status` trailer on the call itself.

Fixes #134

## Test plan
- [x] Added regression test `test_subscribe_to_task_surfaces_jsonrpc_error_on_plain_json_response`
- [x] `cargo test -p a2a-client-lf` (97 passed)
- [x] `cargo clippy -p a2a-client-lf --all-targets`

### DIFF
--- a/a2a-client/src/jsonrpc.rs
+++ b/a2a-client/src/jsonrpc.rs
@@ -1298,7 +1298,10 @@ mod tests {
             tenant: None,
         };
 
-        let error = match transport.subscribe_to_task(&ServiceParams::new(), &req).await {
+        let error = match transport
+            .subscribe_to_task(&ServiceParams::new(), &req)
+            .await
+        {
             Err(e) => e,
             Ok(_) => panic!("expected error, got a stream"),
         };
@@ -1328,7 +1331,10 @@ mod tests {
             tenant: None,
         };
 
-        let error = match transport.subscribe_to_task(&ServiceParams::new(), &req).await {
+        let error = match transport
+            .subscribe_to_task(&ServiceParams::new(), &req)
+            .await
+        {
             Err(e) => e,
             Ok(_) => panic!("expected error, got a stream"),
         };
@@ -1353,7 +1359,10 @@ mod tests {
             tenant: None,
         };
 
-        let error = match transport.subscribe_to_task(&ServiceParams::new(), &req).await {
+        let error = match transport
+            .subscribe_to_task(&ServiceParams::new(), &req)
+            .await
+        {
             Err(e) => e,
             Ok(_) => panic!("expected error, got a stream"),
         };
--- a/a2a-client/src/jsonrpc.rs
+++ b/a2a-client/src/jsonrpc.rs
@@ -1308,6 +1308,61 @@ mod tests {
     }
 
     #[tokio::test]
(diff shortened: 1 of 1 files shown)


muscariello (approved):
LGTM

PR #157, opened 2026-09-03
feat: add a2a-transport-slimrpc CLI plugin crate

Tehsmash:
## Summary

- Adds new `a2acli-transport-slimrpc` crate producing the `a2a-transport-slimrpc` binary
- Implements the pluggable transport contract from [a2aproject/a2a-cli#24](https://github.com/a2aproject/a2a-cli/pull/24), allowing the Go a2a-cli to use SLIMRPC without recompilation
- Drop the binary on PATH; the CLI discovers it automatically as `a2a-transport-slimrpc`

## How it works

The CLI launches `a2a-transport-slimrpc serve --endpoint <slim-url>`. The plugin:

1. Loads config from `$A2A_SLIMRPC_PLUGIN_CONFIG` (YAML file)
2. Connects to the SLIM gateway via `slim_config::ClientConfig`
3. Creates a `SlimRpcTransport` using the configured app identity
4. Starts a TLS gRPC loopback proxy on `127.0.0.1:0` (random port)
5. Writes a handshake JSON line to stdout with the address, token, and self-signed cert PEM
6. Runs until stdin closes (CLI parent exit signal)

The `info` subcommand prints plugin metadata for `a2a transport list`.

## Config file schema

```yaml
client:
  endpoint: "grpc://slim-gateway:46357"
  # full slim_config::ClientConfig (gateway auth, TLS, backoff, etc.)

app:
  name: "org/namespace/agent"
  identity_provider:
    type: shared_secret
    id: "my-id"
    data: "secret"
  identity_verifier:
    type: shared_secret
    id: "my-id"
    data: "secret"
```

### DIFF
--- a/a2acli-transport-slimrpc/src/info.rs
+++ b/a2acli-transport-slimrpc/src/info.rs
@@ -1,0 +1,27 @@
+// Copyright AGNTCY Contributors (https://github.com/agntcy)
+// SPDX-License-Identifier: Apache-2.0
+
+use serde::Serialize;
+
+#[derive(Serialize)]
+pub struct Info {
+    pub name: &'static str,
+    pub version: &'static str,
+    pub description: &'static str,
+    pub protocol: &'static str,
+    pub binding: &'static str,
+}
+
+pub fn run() -> Result<(), crate::error::PluginError> {
+    let info = Info {
+        name: "slimrpc",
+        version: env!("CARGO_PKG_VERSION"),
+        description: "SLIMRPC transport plugin for a2a-cli",
+        protocol: a2a::VERSION,
+        binding: a2a::TRANSPORT_PROTOCOL_GRPC,
+    };
+    let json = serde_json::to_string(&info)
+        .map_err(|e| crate::error::PluginError::Handshake(e.to_string()))?;
+    println!("{json}");
+    Ok(())
+}
--- a/a2acli-transport-slimrpc/src/main.rs
+++ b/a2acli-transport-slimrpc/src/main.rs
@@ -1,0 +1,59 @@
+// Copyright AGNTCY Contributors (https://github.com/agntcy)
+// SPDX-License-Identifier: Apache-2.0
+
+mod error;
+mod info;
+mod serve;
+mod tls;
+
+use clap::{Parser, Subcommand};
+
+#[derive(Debug, Parser)]
+#[command(
+    name = "a2a-transport-slimrpc",
+    version,
+    about = "SLIMRPC transport plugin for a2a-cli"
+)]
+struct Cli {
+    #[command(subcommand)]
+    command: Command,
+}
+
+#[derive(Debug, Subcommand)]
+enum Command {
+    /// Start the loopback proxy server for a given SLIMRPC endpoint.
+    Serve {
+        /// SLIMRPC upstream endpoint (e.g. slim://org/namespace/agent).
+        #[arg(long)]
+        endpoint: String,
+    },
+    /// Print plugin metadata as JSON and exit.
+    Info,
+}
+
+#[tokio::main]
+async fn main() {
+    tracing_subscriber::fmt()
+        .with_env_filter(
+            tracing_subscriber::EnvFilter::try_from_default_env()
+                .unwrap_or_else(|_| "warn".into()),
+        )
+        .with_writer(std::io::stderr)
+        .init();
+
+    rustls::crypto::aws_lc_rs::default_provider()
+        .install_default()
+        .ok();
+
+    let cli = Cli::parse();
+
+    let result = match cli.command {
+        Command::Serve { endpoint } => serve::run(&endpoint).await,
+        Command::Info => info::run(),
+    };
+
+    if let Err(e) = result {
+        eprintln!("a2a-transport-slimrpc: {e}");
+        std::process::exit(1);
+    }
+}
--- a/a2acli-transport-slimrpc/src/serve.rs
+++ b/a2acli-transport-slimrpc/src/serve.rs
@@ -1,0 +1,375 @@
+// Copyright AGNTCY Contributors (https://github.com/agntcy)
+// SPDX-License-Identifier: Apache-2.0
+
+use std::io::{self, BufRead};
+use std::net::SocketAddr;
+use std::sync::Arc;
+
+use a2a::*;
+use a2a_client::Transport;
+use a2a_client::transport::ServiceParams;
+use a2a_grpc::GrpcHandler;
+use a2a_pb::proto::a2a_service_server::A2aServiceServer;
+use a2a_server::RequestHandler;
+use a2a_slimrpc::{SlimRpcTransport, parse_slimrpc_target};
+use async_trait::async_trait;
+use futures::stream::BoxStream;
+use serde::{Deserialize, Serialize};
+use slim_config::auth::identity::{IdentityProviderConfig, IdentityVerifierConfig};
+use slim_config::client::ClientConfig;
+use slim_datapath::api::ProtoName;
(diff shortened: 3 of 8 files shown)


a2acli-transport-slimrpc/src/info.rs
```diff
@@ -0,0 +1,27 @@
+// Copyright AGNTCY Contributors (https://github.com/agntcy)
+// SPDX-License-Identifier: Apache-2.0
+
+use serde::Serialize;
+
+#[derive(Serialize)]
+pub struct Info {
+    pub name: &'static str,
+    pub version: &'static str,
```
msardara, line 9: I'd also add the commit and the build date here

a2acli-transport-slimrpc/src/serve.rs
```diff
@@ -0,0 +1,375 @@
...
+async fn wait_stdin_close() {
+    tokio::task::spawn_blocking(|| {
+        let stdin = io::stdin();
+        let mut reader = stdin.lock();
+        let mut buf = Vec::new();
+        let _ = reader.read_until(0, &mut buf);
+    })
+    .await
+    .ok();
+}
```
msardara, line 375: Is it not possible here to use some tokio native types? Something like

```
use tokio::io;

async fn wait_stdin_close() {
    let stdin = io::stdin();
    let mut reader = io::BufReader::new(stdin);
    let mut buf = Vec::new();
    
    // Reads until a null byte (0) or EOF
    let _ = reader.read_until(0, &mut buf).await;
}
```

a2acli-transport-slimrpc/src/main.rs
```diff
@@ -0,0 +1,59 @@
...
+        /// SLIMRPC upstream endpoint (e.g. slim://org/namespace/agent).
+        #[arg(long)]
+        endpoint: String,
+    },
+    /// Print plugin metadata as JSON and exit.
+    Info,
+}
+
+#[tokio::main]
+async fn main() {
```
msardara, line 35: We could also add it later, but I'd add anyhow to have a nice way to handle errors when they happen.

a2acli-transport-slimrpc/src/serve.rs
```diff
@@ -0,0 +1,375 @@
...
+            .send_streaming_message(&self.forward_params(params), &req)
+            .await
+    }
+
+    async fn get_task(
+        &self,
+        params: &ServiceParams,
+        req: GetTaskRequest,
+    ) -> Result<Task, A2AError> {
+        self.check_token(params)?;
```
msardara, line 157: Are we checking the tokens on each method?

I think a better solution could be to use a tonic interceptor: https://github.com/grpc/grpc-rust/blob/master/examples/src/authentication/server.rs#L50

msardara (commented):
Thanks for your PR @Tehsmash!

Just a few minor comments from my side, and then we should be good to go 🚀

muscariello:
Rebased onto current \`main\` (83 commits behind) via the update-branch API — merged cleanly, no conflicts — and pushed a follow-up commit fixing what CI was catching:

- \`PluginError::UnsupportedBinding\` was never constructed (the binding is hardcoded to gRPC in both \`serve.rs\` and \`info.rs\`, with no negotiation path that could ever produce it), which \`check-compile\`'s denied-warnings policy turns into a hard failure. Removed rather than suppressed, since nothing calls it.
- \`cargo fmt\` had drifted in a few places.
- All 5 new files were missing the required \`Copyright A2A Contributors\` header line.

Verified against the rebased tip: \`just check-compile\`, \`just lint\`, and \`just test\` all pass clean across the full workspace feature matrix. Should be green now.

muscariello:
Pushed a follow-up addressing the 0% patch coverage: extracted \`check_token\`/\`forward_params\`/\`load_config\` out of \`TransportHandler\`/\`run\` as free functions (so they're testable without a live SLIM connection), and added tests for token validation, header forwarding, config loading (valid/missing/malformed YAML/schema mismatch), app-name parsing, the handshake JSON wire shape, TLS cert generation, the info payload, and CLI arg parsing.

Coverage on this crate goes from 0% to ~62%. What's left is genuinely integration-level — \`run()\`'s SLIM connect/serve orchestration and \`main()\`'s process setup — neither reachable from a unit test without a real gateway or a process harness, so left as-is rather than mocked for the sake of a number.

\`just check-compile\`, \`just lint\`, and \`just test\` all still pass clean across the full workspace matrix.

muscariello:
Pushed a follow-up on top of the previous test-coverage commit, since 62% was still low.

**Most of it turned out to be testable, not something to bypass.** `serve.rs` mixed pure logic (config parsing, token checks, handshake shape — already tested) with `TransportHandler`'s dispatch layer and the actual SLIM-gateway-connect/serve-loop orchestration, both untested. Split it into:
- `serve/mod.rs` — the pure logic and its existing tests, unchanged.
- `serve/connect.rs` — `TransportHandler` + `run()`.

`TransportHandler` was hardcoded to `SlimRpcTransport`, so testing its dispatch logic seemed to require a live SLIM connection. Making it generic over the `Transport` trait let it be tested against a fake transport instead — no live connection needed. That raised `serve/connect.rs` from 0% to **~85%** line coverage (25 new tests covering all 10 `RequestHandler` methods: token-rejection and successful forward-and-return for each).

The only part left genuinely untestable in-process is `run()`'s live SLIM gateway connect + TLS/TCP bind + serve loop — no fake-able seam, and it's exercised end-to-end by the itk/csit integration suites instead. Added a root `codecov.yml` (none existed before) that excludes only `main.rs` (its untested lines are `std::process::exit` wiring, which can't be unit-tested without killing the test runner) plus the standard generated-code path.

Crate total is now ~81% covered locally (`cargo llvm-cov -p a2acli-transport-slimrpc`). All of `just check-compile`, `just lint`, `just fmt-check`, `just check-headers`, and `just test` pass for the full workspace.

a2acli-transport-slimrpc/src/info.rs
muscariello: Added in fd64377 — `info` now includes `commit` and `buildDate`, captured at build time via a new `build.rs` (`git rev-parse`/`git log`, falling back to "unknown" if `.git` isn't available, e.g. building from a published tarball).

a2acli-transport-slimrpc/src/serve.rs
muscariello: Done in fd64377 — switched to `tokio::io::BufReader` + `AsyncBufReadExt::read_until` instead of `spawn_blocking` + `std::io::stdin()`.

a2acli-transport-slimrpc/src/serve.rs
muscariello: Good catch — done in fd64377. Moved the check into a tonic `Interceptor` (`check_token_interceptor`), wired via `A2aServiceServer::with_interceptor` in `run_with_config`. It runs once per call, before the request ever reaches `TransportHandler`, which no longer holds a token at all — it only strips it before forwarding upstream. Rewrote the affected tests to exercise the interceptor directly (constructing a `tonic::Request<()>` and checking its metadata) instead of the old per-method ServiceParams checks.

a2acli-transport-slimrpc/src/main.rs
muscariello: Left this as-is for now. `PluginError` (thiserror) already gives clean, typed `Display` messages for every failure mode, and the tests lean on matching specific variants (e.g. `matches!(err, PluginError::InvalidEndpoint(_))`) throughout the crate — `anyhow` would type-erase that at the boundary where it's introduced without changing what actually gets printed to the user. Happy to revisit if there's a specific ergonomics gap it would solve here.

lgecse (approved):

PR #163, opened 2026-09-10
feat: implement ACTS SUT behaviors for the ITK agent with YAML config

JakubWorek:
# Description

SUT for ACTS conformance tests, mirrors [python](https://github.com/a2aproject/a2a-python/pull/1230)

Based on spec: https://github.com/a2aproject/a2a/pull/1882


muscariello (approved):
Reviewed and verified locally. Approving.

Worth recording why the verification was done by hand: `ci.yml` carries `paths-ignore: ['itk/**']` and every `justfile` recipe passes `--exclude itk-rust-current-agent`, so **CI never builds, tests, lints or covers this crate**. The ten green checks on this PR are green because they skipped it entirely. What I ran instead, on the branch rebased onto current `main`:

| | |
| --- | --- |
| `cargo build -p itk-rust-current-agent` | clean |
| `cargo clippy -p itk-rust-current-agent --all-targets -- -D warnings` | clean |
| `cargo fmt --all -- --check` | clean |
| `cargo test -p itk-rust-current-agent` | **17 passed** (`main` has 0) |
| rebase onto `main` | clean |

(The crate needs the `a2a-itk` protos for `itk/build.rs` to run at all — a pre-existing requirement, not something this PR introduces.)

Risk is low: the change is confined to `itk/`, `acts/` and `.gitignore`, so nothing we publish is touched. The worst case is a broken conformance harness, caught by the nightly rather than by a consumer.

The two SDK limitations this works around are real, and I have verified and filed both rather than leave the workarounds undocumented:

- **#203** — `with_capabilities` is write-only. `self.capabilities` is assigned at `handler.rs:459/470/475` and never read, so declared capabilities gate nothing; what actually gates push is whether a store was supplied. Your comment — *"`with_capabilities` is inert in this SDK … withholding the store does"* — is exact, and a harness having to withhold a store to simulate an absent capability is the clearest signal the flag is unwired.
- **#204** — the extended card cannot be configured on `DefaultRequestHandler`, and the not-configured path returns `UNSUPPORTED_OPERATION` (-32004) where A2A defines `EXTENDED_CARD_NOT_CONFIGURED` (-32007). The constant exists in `a2a/src/errors.rs:19` and is mapped in `rest.rs:490`, but nothing emits it.

Both are being worked on now. Once they land, the `acts::capabilities()` gate and the extended-card wrapper here can be reworked to go through the handler — which is the point of fixing them centrally rather than per embedder.

One thing to raise separately, not a blocker for this PR: `paths-ignore: ['itk/**']` made sense when `itk` was a thin harness. It is now 882 lines with 17 tests that nothing gates on a PR, so this will rot silently. A `cargo test -p itk-rust-current-agent` job — it needs `protoc` plus the protos clone, both of which the nightly already provisions — would be cheap insurance.

Nice work, and thanks for the comments explaining the *why* rather than just the what; they are what made the two SDK bugs findable.

PR #171, opened 2026-09-11
feat(a2acli)!: align command surface with the a2a-cli taxonomy

muscariello:
## Summary

Renames `a2acli`'s command surface to match the namespaced taxonomy in the [a2a-cli specification](https://github.com/a2aproject/a2a-cli/blob/main/specification/SPEC.md) §7.1 / Appendix A, closing #165 (part of the Tier 1 conformance tracking issue #164).

| Before | After |
| --- | --- |
| `card` | `card get` |
| `extended-card` | `card get --extended` |
| `send` / `stream` | `send` / `send --stream` |
| `get-task` | `task get` |
| `cancel-task` | `task cancel` |
| `list-tasks` | `task list` |
| `subscribe` | `task subscribe` |
| `push-config ...` | `task push-config ...` |

All `task` operations (including push-config CRUD) now nest under one `task` subcommand, and `card`/`extended-card` collapse into `card get [--extended]` — matching the spec's principle that a command acting on a protocol resource is namespaced by that resource.

## Deliberate deviation from the tracking issue

#165 originally also called for renaming the installed binary from `a2acli` to the upstream Go reference's `a2a`. I did **not** do that: it would let this crate silently shadow the real Go `a2a` binary on any `PATH` where both are installed — a worse outcome than a naming mismatch, and not something graded by any `COMPLIANCE.md` requirement ID. The binary stays `a2acli`.

## Testing

- `cargo test -p a2a-cli` — 19 unit tests + 5 e2e tests, all passing
- `cargo hack clippy -p a2a-cli --all-targets --each-feature -- -D warnings` — clean across `--no-default-features`, `default`, `native-tls`, `rustls-tls`
- `cargo fmt --all -- --check` — clean
- `cargo build --workspace` — clean
- Manually inspected `--help` output at each level (`a2acli --help`, `card --help`, `task --help`, `send --help`) to confirm the new tree

## Not in scope here

Everything else on the Tier 1 checklist in #164 — blocking-by-default `send`/polling (#166), resume hints (#167), `text` output mode and error envelope (#168), auth/transport flags (#169), config precedence (#170).

Issue #170, opened 2026-09-11
a2acli: overridable defaults, config precedence, and config show


muscariello:
## Summary

`SPEC.md` §6.5 requires every baseline default (transport order, wait-by-default, `text` output, concise detail, protocol version, TLS-on) to be overridable by an explicit flag, and §8.3 lets those defaults additionally come from an environment variable or a `.env` file in a fixed precedence (flag > env > local `.env` > global `~/.config/a2a-cli/.env` > built-in), inspectable via a read-only `config show`. `a2acli` has no config-file support, no precedence chain beyond clap's single `env = "A2A_BEARER_TOKEN"` attribute on one flag, and no `config show` command.

## Why

This is what makes repeated invocations short and lets a caller scope settings by agent, per §8.3. It's also the completion of `DEFAULT_001`'s breakdown table (`COMPLIANCE.md` §4a), which needs every default reported as shipped-and-overridable.

## Requirement IDs closed

`A2ACLI_CONFIG_001`, `A2ACLI_DEFAULT_001`

## Scope / plan

- [ ] Add `.env` (dotenv) loading: global `~/.config/a2a-cli/.env`, then a local `.env` discovered by walking up from the working directory, overridden by an explicit `--config <path>` (`CONFIG_001`)
- [ ] Name every environment variable `A2ACLI_<FLAG>` per §7.2 (e.g. `A2ACLI_BEARER`, `A2ACLI_TRANSPORT`, `A2ACLI_CONTEXT_ID`) and wire the full precedence: flag > env > local file > global file > built-in default (`CONFIG_001`)
- [ ] Exclude `--stream`, `-h/--help`, `-v/--version` from environment/file resolution — these remain explicit-flag-only per §7.2
- [ ] Add a `config show` command: read-only, prints each effective setting and the source it resolved from, redacting credential values (`CONFIG_001`)
- [ ] Ensure any persisted secret file is created mode `0600` (or platform equivalent) (`CONFIG_001`)
- [ ] Fill in the `DEFAULT_001` breakdown (transport order / wait-by-default / text output / concise detail / protocol version / TLS-on) — each shipped and overridable by flag — once the send/poll, output-contract, and auth/transport sub-issues land (`DEFAULT_001`)

## Acceptance criteria

- `a2acli config show` prints every effective setting and its source (flag/env/local file/global file/default), with credentials redacted
[...]

Issue #169, opened 2026-09-11
a2acli: auth, transport selection, and version-negotiation flags


muscariello:
## Summary

`SPEC.md` §12–§13 define scriptable, non-interactive credentials (`--bearer`/`--api-key` with env-var equivalents), a general-purpose `--svc-param`, an ordered/repeatable `--transport` preference, `--insecure` with a mandatory warning, and `--debug` with mandatory credential redaction. `a2acli` today has `--bearer-token` (one credential flag, with a non-conforming env var name `A2A_BEARER_TOKEN` instead of `A2ACLI_BEARER`), a generic `--header NAME:VALUE`, and a single-value `--binding <jsonrpc|http-json>` — even though the underlying `a2a-client::A2AClientFactory::preferred_bindings` already takes an ordered `Vec<String>` (`a2a-client/src/factory.rs`). There is no `--api-key`, no `--insecure`, no `--debug`, and therefore no redaction to verify.

## Why

Tier 1 auth/transport requirements are mostly about exposing capability the client library already has, plus a couple of missing flags (`--api-key`, `--insecure`, `--debug`) and the mandatory security warnings/redaction that go with them.

## Requirement IDs closed

`A2ACLI_AUTH_001`–`004`, `A2ACLI_TX_001`–`003`, `A2ACLI_VER_001` (verify only — already ✅ via `a2a-client`), `A2ACLI_CARD_GET_001`

## Scope / plan

- [ ] Rename `--bearer-token` → `--bearer` (keep `--api-key` as a second, independent credential flag) with env-var equivalents named `A2ACLI_BEARER` / `A2ACLI_API_KEY` per §7.2's naming convention (`AUTH_001`, `AUTH_002`)
- [ ] Rename `--header NAME:VALUE` → `--svc-param k:v`, documented as general-purpose and never as an authentication flag (`AUTH_001`)
- [ ] Add `--insecure`, wired to disable TLS verification on the negotiated transport, and emit a mandatory stderr warning whenever a credential is sent over it (`AUTH_003`, plus the `DEFAULT_001` transport-security row)
- [ ] Add `--debug` (raw wire logging to stderr) with credential redaction that cannot be defeated by any verbosity flag (`AUTH_004`) — depends on the output-contract sub-issue's stdout/stderr separation
- [ ] Change `--binding` to `--transport <binding>`, repeatable and ordered (highest preference first), passed straight through to `A2AClientFactory::preferred_bindings` (`TX_001`, `TX_002`)
[...]

Issue #168, opened 2026-09-11
a2acli: text output mode, error envelope, and exit-code contract


muscariello:
## Summary

`SPEC.md` §11 makes `text` the **default**, pipe-safe, human-readable output (`Label: value` per line, no control sequences), with `-o json` as a peer format whose cardinality follows `--stream` (single document vs. JSONL). `a2acli` only ever emits JSON (`print_json` in `a2acli/src/lib.rs`, toggled by `--compact`) — there is no `text` mode. Failures print a plain string to stderr (`main.rs`) rather than the spec's JSON error envelope (`{"error":{"code","message","hint","a2aCode"}}`, Appendix B), and every non-clap-usage failure exits `1` uniformly — there's no distinction for unreachable-agent / auth / timeout conditions, and no `A2ACLI_ERR_*` codes at all (Appendix D).

## Why

This is the single largest Tier 1 gap: `OUT_005` (the text floor) and the interaction-state / resume-hint work both depend on a rendering layer that doesn't exist yet.

## Requirement IDs closed

`A2ACLI_OUT_001`–`006`, `A2ACLI_EXIT_001`–`002`, `A2ACLI_CLI_001` (effective-defaults display)

## Scope / plan

- [ ] Add `-o/--output <text|json>` with `text` as the default (`OUT_005`, `DEFAULT_001`)
- [ ] Implement the `text` renderer: one `Label: value` per line, stable labels across invocations, no terminal control sequences, block-content form (own `Label:` line, content, closing blank line) for rendered artifacts/data parts (`OUT_005`)
- [ ] Keep `-o json` (no `--stream`) as exactly one document — the terminal protocol object, never an event log (`OUT_002`)
- [ ] Emit `-o json --stream` as JSONL — one flushed object per line, final line carrying the terminal object (`OUT_003`)
- [ ] Route all diagnostics/logs to stderr, payload only to stdout, never mixed (`OUT_001`)
- [ ] Replace the plain-string error printer in `main.rs` with the Appendix B error envelope; populate `hint` where actionable (`OUT_004`)
- [ ] Introduce `A2ACLI_ERR_{USAGE,CARD_NOT_FOUND,CARD_INVALID,UNREACHABLE,CREDENTIALS_MISSING,AUTH_FAILED,TIMEOUT,INTERNAL}` per Appendix D, mapped from `CliError` variants; never invent codes outside this namespace (`OUT_004`)
[...]

PR #172, opened 2026-09-11
feat(a2acli): blocking-by-default send, task polling, and message parts

muscariello:
## Summary

Implements Tier 1 requirements `SEND_001`–`006`, `TASK_GET_001`, `TASK_CANCEL_001`, and `TASK_POLL_001`–`004` from the [a2a-cli specification](https://github.com/a2aproject/a2a-cli/blob/main/specification/SPEC.md) (§6.5, §9.1, §9.3, §10.2, §10.3), closing #166 (part of the Tier 1 conformance tracking issue #164).

> **Stacked on #171** (`a2acli-165-command-surface`, the command-surface rename). This PR's own diff is the second commit only; it'll show as a single commit once #171 merges and this is rebased onto `main`.

### Blocking-by-default send and polling (§6.5, §9.3)

- `send` now blocks until the resulting task reaches a terminal or interrupted state, polling `task get` under the hood; `--async` (alias `--no-wait`) returns the identifiers immediately instead.
- `task get` stays one-shot by default; `--wait` turns it into the same poll loop.
- New global `--poll-interval` (default `2s`) / `--timeout` (default `30s`) flags control the loop; expiry surfaces as a new `CliError::Timeout`.
- `--stream` still consumes the SSE stream to completion when it opens; if opening it fails, `send` falls back to a one-shot send plus polling rather than hanging (`TASK_POLL_004`).

**Deliberate deviation:** the spec lists `--return-immediately` as an OPTIONAL alias for `--async`. This CLI already uses `--return-immediately` for the distinct, existing protocol-level `SendMessageConfiguration.return_immediately` field (a request to the *agent* to return early on queued work — not a client-side "don't poll" instruction). Reusing the name would conflate the two, so it's intentionally not aliased; only `--no-wait` is.

### Message parts (§10.2, `SEND_004`–`006`)

- `send` gains `--text-part`, `--file-part`, `--data-part`, and `--media-type`, repeatable and **interleaved-order-preserving**. `--file-part` inlines a local path as base64 bytes or carries a URL by reference (never fetched); `--data-part` reads JSON from a file, stdin (`-`), or an inline string; `--media-type` binds to the part flag immediately preceding it and is a usage error otherwise.
- The positional `text` argument is now optional shorthand for a single `--text-part`, mutually exclusive with the part flags.

Recovering the interleaved order across distinct repeated flags needs the raw `ArgMatches` (`indices_of`), which `Cli::parse()` alone discards — `clap`'s derived struct fields lose cross-flag ordering. `run()` now takes `(Cli, &ArgMatches)`, and a new `run_args()` entry point does the parse-and-run in one call for both `main.rs` and the test suite (which previously called `Cli::try_parse_from` directly).

## Testing

- Extended the e2e fixture with a task that only settles after N `get_task` calls (and one that never settles), so the poll loop, `--async`, `--wait`, and `--timeout` are exercised against real poll counts instead of a task that's already settled on the first read.
- New unit tests: `parse_duration`, `is_settled`, `resolve_message_parts` (ordering, media-type binding, both usage-error paths).
- New e2e tests: blocking send to settlement, `--async` skip, `task get --wait` to settlement, `task get` without `--wait` not polling, `--timeout` firing, full ordered multi-part `send`, local file part + stdin data part, both part-flag usage errors.
- `cargo test -p a2a-cli` — 26 lib + 14 e2e tests, all passing.
- `cargo hack clippy -p a2a-cli --all-targets --each-feature -- -D warnings` — clean across `--no-default-features`, `default`, `native-tls`, `rustls-tls`.

## Not in scope here

Tracked separately under #164: the `text` output mode and error envelope (#168), resume hints (#167), and the remaining auth/transport/config flags (#169, #170) — output is still JSON-only via the pre-existing `--compact` flag.


a2acli/src/lib.rs
```diff
@@ -281,22 +336,57 @@ pub enum CliError {
...
             if command.stream {
-                let client = resolve_client(&cli).await?;
-                let request = build_send_message_request(command, cli.tenant.clone());
-                let stream = client.send_streaming_message(&request).await?;
-                consume_stream(client, stream, cli.compact).await?;
+                match client.send_streaming_message(&request).await {
+                    Ok(stream) => {
+                        consume_stream(client, stream, cli.compact).await?;
+                    }
+                    Err(_stream_unsupported) => {
```
msardara: I think here we should fallback to send_and_maybe_wait if ` e.code == a2a::error_code::UNSUPPORTED_OPERATION`

a2acli/src/lib.rs
```diff
@@ -389,11 +604,185 @@ async fn run_task_command(cli: &Cli, command: &TaskCommand) -> Result<(), CliErr
...
+        return Err(format!("duration must be a non-negative number: {input}"));
+    }
+
+    let seconds = match unit {
+        "ms" => magnitude / 1000.0,
+        "m" => magnitude * 60.0,
+        _ => magnitude,
+    };
+    Ok(Duration::from_secs_f64(seconds))
+}
```
msardara, line 828: We can use the the `duration-string` crate to convert the string to duration: https://docs.rs/duration-string/0.5.3/duration_string/#example

a2acli/src/lib.rs
```diff
@@ -389,11 +604,185 @@ async fn run_task_command(cli: &Cli, command: &TaskCommand) -> Result<(), CliErr
...
+    }
+
+    if tokens.is_empty() {
+        // No part flags given: the positional text (if any) is shorthand for
+        // a single text part.
+        return Ok(command
+            .text
+            .as_deref()
+            .map(|text| vec![Part::text(text)])
+            .unwrap_or_default());
```
msardara:
```suggestion
        return match command.text.as_deref() {
            Some(text) => Ok(vec![Part::text(text)]),
            None => Err(CliError::InvalidInput(
                "message must have at least one part: pass text, or use \
                 --text-part/--file-part/--data-part"
                    .to_string(),
            )),
        };
```

a2acli/src/lib.rs
```diff
@@ -389,11 +604,185 @@ async fn run_task_command(cli: &Cli, command: &TaskCommand) -> Result<(), CliErr
...
+    }
+    if let Some(indices) = matches.indices_of("data_parts") {
+        tokens.extend(
+            indices
+                .zip(command.data_parts.iter())
+                .map(|(index, value)| (index, Token::Part(Kind::Data, value.as_str()))),
+        );
+    }
+
+    if tokens.is_empty() {
```
msardara, line 670: If a media-type is present, we are silently dropping it when entering this condition. We should probably check that:


```suggestion
    if !command.media_types.is_empty() && tokens.is_empty() {
        return Err(CliError::InvalidInput(
            "--media-type must immediately follow a --text-part/--file-part/--data-part"
                .to_string(),
        ));
    }

if tokens.is_empty() {
```

a2acli/src/lib.rs
```diff
@@ -389,11 +604,185 @@ async fn run_task_command(cli: &Cli, command: &TaskCommand) -> Result<(), CliErr
...
+            .map_err(|source| CliError::ReadFile {
+                path: "<stdin>".to_string(),
+                source,
+            })?;
+        return Ok(Part::data(serde_json::from_str(&buffer)?));
+    }
+
+    if let Ok(text) = std::fs::read_to_string(value) {
+        return Ok(Part::data(serde_json::from_str(&text)?));
+    }
```
msardara:
```suggestion
    match std::fs::read_to_string(value) {
        Ok(text) => return Ok(Part::data(serde_json::from_str(&text)?)),
        Err(source) if source.kind() != std::io::ErrorKind::NotFound => {
            return Err(CliError::ReadFile {
                path: value.to_string(),
                source,
            });
        }
        Err(_) => {} // not a file path — try parsing as inline JSON below
    }
```

msardara (changes requested):
LGTM, just few minor nits to fix ✅

a2acli/src/lib.rs
muscariello: Good catch, fixed exactly as suggested: the fallback to `send_and_maybe_wait` now only fires when `error.code == a2a::error_code::UNSUPPORTED_OPERATION`; any other error opening the stream now propagates instead of being silently retried. Added a test (`send_stream_propagates_a_non_unsupported_error_instead_of_falling_back`) covering the non-fallback case. Pushed.

a2acli/src/lib.rs
muscariello: Thanks for the pointer — I looked at `duration-string`, but decided to keep the hand-rolled parser: it requires a unit suffix on every value (no bare-number-as-seconds shorthand) and integer magnitudes only (no "1.5s"), both of which the current parser supports and are already covered by its unit tests. Didn't seem like a net win to add a dependency and lose that flexibility. Happy to reconsider if you feel strongly about it.

a2acli/src/lib.rs
muscariello: Fixed exactly as suggested — a message with no positional text and no part flags now returns `CliError::InvalidInput` instead of an empty parts array. Added `send_rejects_a_message_with_no_content_at_all`. Pushed.

a2acli/src/lib.rs
muscariello: Good catch, fixed as suggested — a lone `--media-type` (no other part flag at all) is now rejected before the "no part flags given" branch can silently skip it. Added `media_type_alone_with_no_other_part_flag_is_rejected`. Pushed.

a2acli/src/lib.rs
muscariello: Fixed exactly as suggested — only a `NotFound` error now falls through to inline-JSON parsing; any other `std::fs::read_to_string` error (permission denied, a directory, invalid UTF-8, ...) is reported as `CliError::ReadFile` immediately. Added `data_part_reports_the_real_error_for_an_unreadable_existing_path` (using a directory path). Pushed.

msardara (approved):

Issue #164, opened 2026-09-11
a2acli: Tier 1 conformance with the a2a-cli specification


muscariello:
## Summary

[`a2a-cli`](https://github.com/a2aproject/a2a-cli) (Go) defines a standardized, cross-language command-line contract for driving A2A agents: [`specification/SPEC.md`](https://github.com/a2aproject/a2a-cli/blob/main/specification/SPEC.md) (normative) and [`specification/COMPLIANCE.md`](https://github.com/a2aproject/a2a-cli/blob/main/specification/COMPLIANCE.md) (the requirement-identifier registry and acceptance checklist). `a2acli`, the CLI shipped from this repo (`a2acli/`), needs to conform to it.

A review of `a2acli/src/lib.rs` against `SPEC.md` v0.2 / `COMPLIANCE.md` (2026-08-17) found it currently satisfies **none of the 39 Tier 1 ("Core") requirements** cleanly — it's a thin JSON-only wrapper around `a2a-client` with a different command surface, no blocking/polling, and no human-readable output mode. This issue tracks bringing it to Tier 1.

## Why

Tier 1 is the baseline every `a2a-cli` implementation is meant to share (`SPEC.md` §5), so a script, CI job, or AI coding agent written against the spec's command surface behaves identically regardless of implementation language (§6.4). Today a script written against the standard surface (`a2a card get <url>`, `a2a send --stream ...`, `a2a task get <id> --wait`) does not run against `a2acli` at all — different command names, different flags, no blocking wait, JSON-only output.

## Scope

Tracked as 6 sub-issues, each an independent unit of work; each lists the exact `COMPLIANCE.md` §4 requirement IDs it closes.

- [ ] #165 — align command surface with the a2a-cli taxonomy
- [ ] #166 — blocking-by-default `send`, task polling, and message parts
- [ ] #167 — stateless interaction-id handling and resume hints
- [ ] #168 — `text` output mode, error envelope, and exit-code contract
- [ ] #169 — auth, transport selection, and version-negotiation flags
- [ ] #170 — overridable defaults, config precedence, and `config show`

## Requirement tally (Tier 1 — Core, 39 total)

| Area | IDs | Current status |
| --- | --- | --- |
| `CARD_GET` | `_001` | ◐ fetches card, but no `--extended`/ordered transport preference |
| `SEND` | `_001`–`_006` | ◐ sends text only; no blocking wait, no message parts |
[...]

PR #176, opened 2026-09-11
feat(a2acli): auth, transport selection, and version-negotiation flags

muscariello:
## Summary

Implements Tier 1 requirements `AUTH_001`–`004`, `TX_001`–`003`, `VER_001` (verified), and `CARD_GET_001` (confirmed) from the [a2a-cli specification](https://github.com/a2aproject/a2a-cli/blob/main/specification/SPEC.md) (§12, §13), closing #169 (part of the Tier 1 conformance tracking issue #164).

> **Stacked on #174** (`a2acli-167-interaction-state`). This PR's own diff is the last two commits only.

### Auth (§12.1, `AUTH_001`/`002`)

- `--bearer-token` → `--bearer`, with an `A2ACLI_BEARER` env var replacing the non-conforming `A2A_BEARER_TOKEN` name (§7.2's naming convention).
- New `--api-key` (env `A2ACLI_API_KEY`), attached as an `X-API-Key` header. `a2acli` doesn't yet read the agent card's declared security scheme to place it per A2A §4.5.2 — a documented simplification, not something to rely on for a scheme that expects it elsewhere.
- `--header` → `--svc-param`, documented as general-purpose and never an authentication flag, distinct from `--metadata`.

### `--insecure` (§7.2, `AUTH_003`, `DEFAULT_001` transport-security row)

Disables TLS certificate verification for both the card fetch and the negotiated transport, via a custom `reqwest::Client` with `danger_accept_invalid_certs`, injected into `JsonRpcTransportFactory`/`RestTransportFactory`. Gated behind the `rustls-tls`/`native-tls` features — with neither compiled in there's no verification to disable, so the flag is a no-op rather than a build error. **Always** prints a stderr warning, naming the specific risk when a credential is also configured — never disables verification silently.

### `--debug` (§7.2, `AUTH_004`)

Enables `a2a-client`'s existing `LoggingInterceptor` (method-level request/response diagnostics) plus a `tracing-subscriber` writer to stderr. `AUTH_004`'s redaction requirement is satisfied by construction at this scope: the diagnostics never include headers, params, or bodies, so there's no credential-bearing field to leak in the first place. Full wire-level `--debug` content (raw protocol messages) is Tier 2 (`OUT_007`) and out of scope here.

### Transport (§13.1, `TX_001`/`002`)

`--binding <jsonrpc|http-json>` → `--transport`, repeatable and ordered (highest preference first), passed straight through to `A2AClientFactory::preferred_bindings`. The HTTP+JSON binding's CLI value is renamed `http-json` → `rest` to match the reference implementation's `--transport` vocabulary.

### `TX_003` — selected interface's own tenant

Adds `A2AClientFactory::create_from_card_with_interface` to `a2a-client` (additive; `create_from_card`'s existing signature and behavior are unchanged) so the CLI can recover which `AgentInterface` was selected — there was previously no way to get it back once a client had been built. `resolve_client` now resolves the tenant to use on every subsequent request as `cli.tenant`, falling back to the selected interface's own declared tenant when the caller didn't pass `--tenant`.

### Verified, no code change

- `VER_001`: `A2A-Version` is already signaled on every request by `a2a-client` itself (`SVC_PARAM_VERSION` in `client.rs`).
- `CARD_GET_001`: `card get` already uses the resolved card to select the transport for subsequent operations.

## Testing

- `a2a-client`: new `create_from_card_with_interface` test asserting the selected interface (and its tenant) round-trips correctly; `create_from_card`'s own behavior is unchanged and its existing tests still pass.
- `a2acli` e2e: new tests for `--api-key` header attachment, `--insecure`'s warning (with and without a configured credential, and its absence when `--insecure` isn't passed), `--debug` producing diagnostics without the bearer token appearing in them, and both the `TX_003` tenant fallback and its override by an explicit `--tenant`.
- `cargo test -p a2a-client-lf` — 101 tests passing.
- `cargo test -p a2a-cli` — 30 lib + 28 e2e tests passing.
- `cargo hack clippy --all-targets --each-feature -- -D warnings` — clean for both crates across every feature combination (including `--no-default-features`, where `--insecure`'s TLS logic compiles out).

## Not in scope here

Full wire-level `--debug` content (Tier 2, `OUT_007`); reading the agent card's declared security scheme to place `--api-key` per its own rules rather than a fixed `X-API-Key` header; config precedence and `config show` (#170).


msardara (approved):
LGTM ✅

Issue #165, opened 2026-09-11
a2acli: align command surface with the a2a-cli command taxonomy


muscariello:
## Summary

`SPEC.md` §7.1 defines a namespaced command surface (`card get`, `send`, `task get`, `task cancel`, `task list`, `task subscribe`, `task push-config …`) so a caller can predict where an operation lives. `a2acli`'s current surface is flat and named differently:

| Spec | `a2acli` today |
| --- | --- |
| `card get` | `card` |
| `card get --extended` (flag, §10.1) | `extended-card` (separate command) |
| `send` / `send --stream` | `send` / `stream` (separate command) |
| `task get` | `get-task` |
| `task cancel` | `cancel-task` |
| `task list` | `list-tasks` |
| `task subscribe` | `subscribe` |
| `task push-config create/get/list/delete` | `push-config create/get/list/delete` ✅ already matches |

The published binary name also differs: the spec's reference binary is `a2a`; `a2acli/Cargo.toml`'s `[[bin]]` produces `a2acli`.

## Why

None of the numbered Tier 1 requirements are reachable under their spec-mandated names while the command surface diverges this much — a script or agent skill written against the standard `a2a <command>` surface fails outright against `a2acli`. This is foundational; the other Tier 1 issues assume these command names exist.

## Scope / plan

- [ ] Rename `Command::Card` → `card get` (fold `extended-card` into `card get --extended`, per §10.1)
- [ ] Rename `Command::GetTask` → `task get`
- [ ] Rename `Command::CancelTask` → `task cancel`
- [ ] Rename `Command::ListTasks` → `task list`
- [ ] Rename `Command::Subscribe` → `task subscribe`
- [ ] Fold `Command::Stream` into `send --stream` (§7.2, §10.2) rather than a separate top-level command
- [ ] Nest the `task …` commands under a `Task { command: TaskCommand }` subcommand group, alongside the existing `PushConfig` group (or nest push-config under `task push-config` if that's not already the effective shape)
- [ ] Update the `[[bin]]` name in `a2acli/Cargo.toml` (and README/install instructions) to match the spec's reference binary name
- [ ] Update `README.md` and `tests/e2e.rs` for the new invocation forms

## Acceptance criteria

- `a2acli card get <url>`, `a2acli send --stream ...`, `a2acli task get/cancel/list/subscribe ...` all exist and behave as described in `SPEC.md` §7.1 / Appendix A
[...]

PR #177, opened 2026-09-11
feat(a2acli): overridable defaults, config precedence, and config show

muscariello:
## Summary

Implements the final Tier 1 requirements `CONFIG_001` and `DEFAULT_001` from the [a2a-cli specification](https://github.com/a2aproject/a2a-cli/blob/main/specification/SPEC.md) (§6.5, §8.3), closing #170 — **the last of the six sub-issues tracked under #164**.

> **Stacked on #176** (`a2acli-169-auth-transport-version`). This PR's own diff is the last commit only.

### Config precedence (§8.3): flag > environment variable > local `.env` > global `.env` > built-in default

- Added `env = "A2ACLI_<FLAG>"` to every eligible global `Cli` field (`base-url`, `transport`, `insecure`, `debug`, `tenant`, `output`, `compact`, `async`, `wait`, `poll-interval`, `timeout`) plus `send`'s `--context-id`/`--task-id`, matching §7.2's naming convention. `--stream`, `-h/--help`, and `-v/--version` remain explicit-flag-only, as required.
- New `--config <path>` flag: an explicit `.env` file in place of the local `.env` search.
- New `load_dotenv_config`: discovers the global `.env` (`~/.config/a2a-cli/.env`, `$XDG_CONFIG_HOME`-aware) and the local one (walking up from the working directory, like `git`), parses simple `KEY=value` syntax (`#` comments, blank lines, a tolerated `export ` prefix, one layer of stripped quotes), and injects any `A2ACLI_*` key that isn't already a real environment variable — which is what makes clap's own `env = "..."` resolution on `Cli`'s fields see it. A real environment variable always outranks either file, snapshotted before injection so a local file can still override an already-injected global value for the same key.
- `a2acli` never writes a `.env` file itself, so `CONFIG_001`'s "secret files must be mode `0600`" has no file for the tool to create insecurely — but it now warns when a `.env` file it *reads* is readable by users other than its owner.

### `config show` (new `config` command, `CONFIG_001`)

Read-only, prints each effective setting and the source it resolved from (flag / environment variable / local `.env` file / global `.env` file / built-in default), using clap's own `ArgMatches::value_source` plus the dotenv-injection provenance to distinguish a real env var from one this run injected from a file. Credential values (`--bearer`, `--api-key`) are redacted. Renders in both `text` and `-o json` like every other command's output:

```
$ a2acli --bearer secret123 --transport jsonrpc,rest config show
base-url: http://localhost:3000 (source: built-in default)
transport: jsonrpc,rest (source: flag)
bearer: (set, redacted) (source: flag)
...
```

### `DEFAULT_001`

All six of §6.5's baseline defaults were already shipped by earlier work in this effort; this closes the loop by making every one of them additionally overridable via environment/file, completing the full precedence chain the spec requires.

## Testing

- New unit tests for `parse_dotenv` (comments, blank lines, `export` prefix, quote stripping), `find_flag_value` (space and `=` forms), and the `ConfigSource`/redaction helpers `config show` relies on.
- New e2e tests, each using an isolated scratch working directory (with `HOME`/`XDG_CONFIG_HOME` repointed at it) so they don't depend on or interfere with the real developer machine or each other despite running in parallel: a local `.env` setting a default with no flag, a real environment variable outranking a conflicting local `.env` value, `--config` overriding local-file discovery, and `config show` reporting both flag-sourced (redacted) and file-sourced settings correctly.
- `cargo test -p a2a-cli` — 37 lib + 33 e2e tests, all passing.
- `cargo hack clippy -p a2a-cli --all-targets --each-feature -- -D warnings` — clean across `--no-default-features`, `default`, `native-tls`, `rustls-tls`.

## Tracking

This closes out the a2acli Tier 1 conformance effort tracked in #164 — all six sub-issues (#165–#170) now have PRs open: #171 (merged), #172, #173, #174, #176, and this one.


msardara (approved):
LGTM ✅

Issue #238, opened 2026-09-15
Introduce fuzz testing for the protocol parsing surfaces


muscariello:
## Why

Two bugs found by hand this month were in exactly the surfaces a fuzzer covers:

- **#198** — residual bytes at the end of an SSE stream were dropped. A framing
  bug, found by reading the code.
- **#223** — a trailing dot (`localhost.`) bypassed the push URL host blocklist.
  Found by manually enumerating URL spellings against `validate_push_url`.

Both are the classic shapes fuzzing finds cheaply and humans find expensively.
This SDK parses untrusted bytes from the network on every request — JSON-RPC
envelopes, SSE frames, Agent Cards, protobuf, protojson — and none of it is
fuzzed.

Scorecard's `Fuzzing` check is also 0, but that is a byproduct, not the goal.
Its Rust detector is satisfied by **any `*.rs` file containing
`libfuzzer_sys`** (`checks/raw/fuzzing.go`), so a single trivial target would
score 10 while finding nothing. Worth stating plainly so the score does not
become the objective.

No sibling SDK fuzzes either (`a2a-python`, `a2a-go`, `a2a-dotnet` have no fuzz
directory), so there is no in-house pattern to copy and no differential oracle
available off the shelf.

## Constraints

- `cargo-fuzz` requires a **nightly** toolchain and is **Unix-only** (not
  Windows). Our MSRV is stable 1.85 and CI runs the stable matrix across
  ubuntu/macos/windows, so fuzzing needs its own Linux + nightly job, outside
  that matrix.
- `fuzz/` must be excluded from the workspace, or `cargo build --workspace` on
  stable breaks for everyone.
- Consider the CI cache budget: the repo is already at 9.6 GiB of GitHub's
  10 GiB and evicting within hours. A fuzz job must not save a cache on pull
  requests.

## Targets, ranked by evidence rather than guesswork

| # | Target | Why this one |
| --- | --- | --- |
| 1 | `a2a-client::jsonrpc::{parse_sse_bytes, parse_stream_tail, parse_sse_stream_rest}` | #198 proves this code harbours framing bugs. Arbitrary byte splits across frame boundaries is precisely what a fuzzer explores and a test author does not. |
| 2 | `a2a_pb::protojson_conv::{to_value, from_value}` | A genuine **round-trip pair**, plus `transcode_message<Src, Dst>` between protobuf and protojson. Two encodings of one value is a differential oracle that needs no external implementation. |
[...]

PR #266, opened 2026-09-17
fix(a2a-pb): reject out-of-range timestamps in to_proto_timestamp (#261)

### LINKED ISSUE #261: protojson round trip on an out-of-range Timestamp produces unparseable output
## Summary

`a2a_pb::pbconv::to_proto_timestamp` converts a `chrono::DateTime<Utc>` to a
`prost_types::Timestamp` unconditionally:

```rust
pub fn to_proto_timestamp(dt: &chrono::DateTime<chrono::Utc>) -> prost_types::Timestamp {
    prost_types::Timestamp {
        seconds: dt.timestamp(),
        nanos: dt.timestamp_subsec_nanos() as i32,
    }
}
```

`google.protobuf.Timestamp`'s own doc comment states it is only meaningful for
`0001-01-01T00:00:00Z` to `9999-12-31T23:59:59.999999999Z`. `chrono::DateTime`
represents a much wider range, so a value outside that window converts without
error but produces JSON the crate's own protojson deserializer then rejects.

## Reproduction

Found while building fuzz target 2 (#238 Phase 2), fuzzing the `to_value`/
`from_value` round trip on `Task` with `arbitrary`-generated timestamps:

```
Task.status.timestamp = +149307-03-21T11:48:32.907012289Z

to_value(&task)   -> Ok({"status":{"timestamp":"+149307-03-21T11:48:32.907012289+00:00", ...}})
from_value(value) -> Err(ProtoJSON error: input contains invalid characters)
```

`to_value` succeeds and produces a string with a `+`-prefixed six-digit year;
`from_value` on that exact output fails. Same crate, same round trip,
disagreeing with itself — the two encodings should be equivalent (#238's
target 2 rationale) and are not.

## Impact

Low severity: this needs a `Task`/`TaskStatusUpdateEvent`/etc. carrying a
timestamp no real clock produces, so it is not remotely triggerable by an
attacker in the way #223 was. It is nonetheless a genuine gap — nothing
validates the range on the way in, so a task store or agent implementation
that accepts a caller-supplied timestamp without its own bound would silently
produce a card/task that a client's own SDK cannot parse back.

## Fix direction

Not fixed here — this issue is the tracking place for that, and the fuzz
regression corpus (a2a-pb's round-trip fuzz target, #238) carries the
minimal reproducing value so a fix is verified against the exact case that
found it. Two directions, not mutually exclusive:

- Validate the range where a `Task`/event carrying a timestamp is
  constructed or accepted from a caller, rejecting rather than converting.
- Have `to_proto_timestamp` clamp or error rather than pass an out-of-range
  value through silently.

chopmob-cloud:
Fixes #261.

`to_proto_timestamp` converted a `chrono::DateTime<Utc>` to a `prost_types::Timestamp` unconditionally. `google.protobuf.Timestamp` is only defined for `0001-01-01T00:00:00Z` to `9999-12-31T23:59:59.999999999Z`, so a value outside that window round-tripped inconsistently within the same crate: `to_value` produced a `+`-prefixed six-digit-year string and `from_value` on that exact output failed.

Per the issue's fix direction, `to_proto_timestamp` now returns `Option<Timestamp>` and yields `None` outside the representable range, mirroring the already-fallible `from_proto_timestamp`. The two callers move from `.map` to `.and_then`, so an out-of-range status timestamp drops to `None` rather than emitting a `TaskStatus` this crate cannot read back.

Bounds: seconds in `[-62_135_596_800, 253_402_300_799]`.

Tests: a boundary unit test and a `TaskStatus` round-trip regression for the value class fuzz target 2 (#238) found. `cargo test -p a2a-pb`, `cargo fmt --check`, and `cargo clippy -- -D warnings` all pass.


a2a-pb/src/pbconv.rs
```diff
@@ -2042,12 +2061,45 @@ mod tests {
...
+    fn test_to_proto_timestamp_rejects_out_of_range() {
+        // Regression for #261: a chrono value beyond google.protobuf.Timestamp's
+        // range must not silently convert into unparseable protojson.
+        let too_late = chrono::DateTime::from_timestamp(300_000_000_000, 0).unwrap();
+        assert!(to_proto_timestamp(&too_late).is_none());
+        // Both inclusive boundaries still convert.
+        let max = chrono::DateTime::from_timestamp(PROTO_TIMESTAMP_MAX_SECONDS, 0).unwrap();
+        let min = chrono::DateTime::from_timestamp(PROTO_TIMESTAMP_MIN_SECONDS, 0).unwrap();
+        assert!(to_proto_timestamp(&max).is_some());
+        assert!(to_proto_timestamp(&min).is_some());
```
muscariello, line 2080: Nice regression test — one gap: it checks the exact boundaries and a value ~10,000× out of range, but nothing at `MAX+1`/`MIN-1`, exactly where an off-by-one would slip through unnoticed.
```suggestion
        assert!(to_proto_timestamp(&min).is_some());
        assert!(
            to_proto_timestamp(
                &chrono::DateTime::from_timestamp(PROTO_TIMESTAMP_MAX_SECONDS + 1, 0).unwrap()
            )
            .is_none(),
            "one second past the max must still be rejected"
        );
        assert!(
            to_proto_timestamp(
                &chrono::DateTime::from_timestamp(PROTO_TIMESTAMP_MIN_SECONDS - 1, 0).unwrap()
            )
            .is_none(),
            "one second before the min must still be rejected"
        );
```

muscariello (commented):
The fix itself is correct — I independently recomputed both range constants (`-62_135_596_800` / `253_402_300_799`) via `chrono` myself, they match exactly, and every call site (`to_proto_task_status`, `to_proto_list_tasks_request`) is consistently updated from `.map` to `.and_then`. This correctly closes #261 and nothing here blocks merging.

A few things worth addressing:

**1. `from_proto_timestamp` (unchanged) is now the more permissive half of the pair.**
The new doc comment says `to_proto_timestamp` "mirrors `from_proto_timestamp`, which is likewise fallible" — but `from_proto_timestamp`'s fallibility comes entirely from `chrono`'s own (much wider, ~±262,000-year) range, not from `google.protobuf.Timestamp`'s declared one. A wire-received `Timestamp` far outside the protobuf-valid range is accepted into a `Task` with no complaint, then silently drops the next time it's converted back out via the newly-strict `to_proto_timestamp`. Might be worth softening that doc claim, or filing a follow-up to bound `from_proto_timestamp` symmetrically.

**2. Could you remove the now-inaccurate part of the comment on `a2a/src/types.rs:23-26`?** It currently says *"a2a-pb's to_proto_timestamp still converts the value... a genuine round-trip bug, filed as #261 rather than fixed here"* — since this PR is that fix, it'll describe an already-closed bug as open the moment it merges.
Reduce comments verbosity please.

**3. Could you retitle this PR (and its commit) to `fix(a2a-pb)!:`?** `to_proto_timestamp` is `pub fn` in `pub mod pbconv`, and its return type changing from `Timestamp` to `Option<Timestamp>` is source-breaking for any external caller. This repo marks that with `!` even pre-1.0 (e.g. `fix(a2a-server)!:` for a similarly-shaped change, recorded `[**breaking**]` in this crate's own CHANGELOG.md) — without it, `release-plz` will cut a silent patch bump instead of flagging the break.

muscariello (changes requested):
Can you please check the suggestions to apply ?

muscariello (approved):
LGTM

muscariello:
@chopmob-cloud thanks for your contribution

PR #267, opened 2026-09-17
fix(a2a-server): screen resolved push-URL addresses at connect time (#224)

### LINKED ISSUE #224: a2a-server: push URL SSRF guard does not resolve hostnames, so CWE-918 is still open
## Summary

`validate_push_url` (added in #124) inspects the **URL string** at parse time.
It never resolves the hostname, so any DNS name pointing at a restricted
address passes:

```
block  http://10.0.0.5/x
block  http://169.254.169.254/x
ALLOW  http://internal.corp.test/x   <-- resolves wherever DNS says
```

Confirmed against `main` at `2ba5017`.

This leaves CWE-918 open in the general case:

- **Internal names.** Any `*.internal`, `*.corp`, or split-horizon name
  resolving to RFC 1918 space reaches the service. The blocklist cannot see it.
- **Attacker-controlled DNS.** A public name whose A record is `127.0.0.1` or
  `169.254.169.254` passes validation and connects where the attacker chose.
- **DNS rebinding.** Even a check that resolved once at validation time would
  be racing the connection: the name can resolve differently when `reqwest`
  dials (TOCTOU).

## Why this is not a defect in #124

The PR description cites Go's `ssrfGuardedTransport` + `isBlockedIP` as
precedent, but those differ in kind. Go guards the **transport**, checking the
resolved peer address at connect time, which is immune to all three cases
above. String validation is a genuinely useful first line — it stops the
obvious literal-address attempts, and #124 is a clear improvement on having
nothing — but the two are not equivalent and the docs should not imply they
are.

## Scope

- [ ] Guard at connect time on the **resolved** address rather than the
      hostname, so the check cannot be separated from the dial. With `reqwest`
      this means a custom connector / resolver that applies the existing
      `is_blocked_ip` logic to each candidate socket address before connecting.
- [ ] Reuse the range predicates already written for #124 rather than
      duplicating them — loopback, RFC 1918, link-local, IPv6 ULA,
      unspecified, multicast, IPv4-mapped IPv6.
- [ ] Keep the string check as a cheap pre-filter; it gives a clear error
      before any network activity.
- [ ] Honour `HttpPushSenderConfig::validate_urls` for the connect-time guard
      too, so local test webhooks keep their single opt-out.
- [ ] Correct the claim of parity with Go in the docs and release notes, and
      state what the string check does and does not cover.

## Related

- #124 — introduced `validate_push_url`.
- #223 — a concrete bypass of the string check itself (trailing dot). Separate
  and much smaller; fixing that does not address anything here.

chopmob-cloud:
Fixes #224.

`validate_push_url` inspects the URL string at parse time, so a hostname that resolves to a loopback, private, link-local, or metadata address still passed (split-horizon DNS, attacker-controlled A records, DNS rebinding). That left CWE-918 open in the general case, as #224 sets out.

This moves the guard to connect time and closes the paths that reach an address the string check never saw. All three are gated on `validate_urls`, so the single local-webhook opt-out still disables the whole guard:

- **Resolver:** a `reqwest` DNS resolver (`SsrfGuardedResolver`) runs the existing `is_blocked_ip` predicate over every resolved `SocketAddr` and drops the blocked ones before the socket is dialed, erroring when none remain. Because reqwest dials exactly the addresses the resolver returns, the check cannot be separated from the connect, so there is no resolve-then-reconnect TOCTOU.
- **Redirects:** a custom redirect policy re-screens each hop with `validate_push_url`. A redirect to a blocked literal IP skips DNS entirely (so it never reaches the resolver) and the string check only ran on the original URL, so without this a `public -> http://127.0.0.1/` redirect would bypass the guard.
- **Proxy:** `.no_proxy()` under validation, since a configured system proxy would resolve and reach the target itself, leaving only the proxy address screened.

Per the issue's scope, the resolver reuses `is_blocked_ip` rather than duplicating the range predicates, and the string check is kept as a cheap pre-filter that fails before any network activity.

Tests: `connectable_addrs` drops blocked and keeps public addresses; an all-blocked set yields empty; the resolver refuses a host resolving only to loopback; and the redirect-hop check refuses blocked literal targets. `cargo test -p a2a-server-lf`, `cargo fmt --check`, and `cargo clippy -- -D warnings` all pass.

Two notes for reviewers:
- The `is_blocked_ip` predicate reused here (from #124) covers loopback, RFC 1918, link-local, ULA, IPv4-mapped IPv6, unspecified, and multicast, but not CGNAT `100.64.0.0/10`, the IPv4 broadcast address, or `192.0.0.0/24`. Extending it is a separate change from this one (which is about where the check runs); flagging it so it is a conscious decision rather than a silent gap.
- PR #124 described string validation as equivalent to Go's connect-time transport guard; they differ in kind. The doc comments here state what each layer does (string pre-filter, resolved-address screening, per-hop redirect re-screening, proxy disabled) so the distinction is explicit.


muscariello (commented):
Verified this fixes #224 end to end (connect-time resolver, redirect re-screening, proxy disabled, all gated on `validate_urls`, string check kept as a pre-filter). The only real gap I found is the codecov/patch failure: the redirect-policy closure is genuinely untested. Not blocking (it's not in the required status checks), but here's a suggestion to close it with a real test rather than a synthetic one -- `reqwest::redirect::Attempt` has no public constructor, so the closure has to be pulled out and driven through an actual local-server redirect to be exercised at all.

a2a-server/src/push/sender.rs
```diff
@@ -39,10 +49,37 @@ impl Default for HttpPushSenderConfig {
...
+                .redirect(reqwest::redirect::Policy::custom(|attempt| {
+                    if attempt.previous().len() >= 10 {
+                        return attempt.stop();
+                    }
+                    if validate_push_url(attempt.url().as_str()).is_err() {
+                        attempt.error("push URL redirect target is blocked")
+                    } else {
+                        attempt.follow()
+                    }
+                }));
```
muscariello: This closure is the one piece of the redirect guard that's completely untested — `reqwest::redirect::Attempt` has no public constructor, so nothing here can invoke it directly, and no test drives a real HTTP redirect through it either. That's exactly what's showing up as the 9 missing patch-coverage lines.

Pulling it out into a named function lets a test attach the *same* policy to a bare `reqwest::Client` (no resolver, no proxy override) and drive it with a real local server that issues an actual redirect — closing the gap on the part that matters (does this actually block a bad redirect) rather than testing the logic in isolation.

```suggestion
                .redirect(ssrf_guarded_redirect_policy());
```

a2a-server/src/push/sender.rs
```diff
@@ -230,6 +267,51 @@ pub(crate) fn is_blocked_ip(ip: std::net::IpAddr) -> bool {
...
+                return Err(Box::<dyn std::error::Error + Send + Sync>::from(
+                    "push URL host resolves only to blocked (loopback, private, \
+                     link-local, or metadata) addresses",
+                ));
+            }
+            Ok(Box::new(allowed.into_iter()) as Addrs)
+        })
+    }
+}
+
```
muscariello, line 305:
```suggestion

/// The redirect policy installed when `validate_urls` is set: re-screens
/// each hop with [`validate_push_url`], since a redirect to a blocked
/// literal IP skips DNS entirely and never reaches [`SsrfGuardedResolver`].
///
/// Split out from [`HttpPushSender::new`] so a test can attach it to a
/// client on its own, without the resolver -- `reqwest::redirect::Attempt`
/// has no public constructor, so driving a real HTTP redirect through this
/// policy is the only way to exercise it.
fn ssrf_guarded_redirect_policy() -> reqwest::redirect::Policy {
    reqwest::redirect::Policy::custom(|attempt| {
        if attempt.previous().len() >= 10 {
            return attempt.stop();
        }
        if validate_push_url(attempt.url().as_str()).is_err() {
            attempt.error("push URL redirect target is blocked")
        } else {
            attempt.follow()
        }
    })
}

```

a2a-server/src/push/sender.rs
```diff
@@ -626,4 +708,65 @@ mod tests {
...
+    fn test_redirect_hop_check_rejects_blocked_targets() {
+        // The custom redirect policy re-runs validate_push_url on each hop, so a
+        // redirect to a blocked literal IP (which skips DNS and never reaches
+        // SsrfGuardedResolver) is refused, while a public hop follows. This pins
+        // the per-hop check the policy relies on.
+        assert!(validate_push_url("http://127.0.0.1/").is_err());
+        assert!(validate_push_url("http://169.254.169.254/latest/meta-data/").is_err());
+        assert!(validate_push_url("http://[::1]/").is_err());
+        assert!(validate_push_url("https://example.com/next").is_ok());
+    }
```
muscariello, line 783: Two tests, both using a real local server so they exercise the actual policy through a genuine `reqwest::redirect::Attempt`:

- a redirect to a disallowed target (169.254.169.254) is refused
- a server that redirects to itself (loopback) is refused on the first hop, rather than followed

Both are self-contained and deterministic — no external network dependency. I couldn't find a way to also cover the `attempt.follow()` / 10-hop-cap branches without one: any redirect target a local test controls is itself loopback, so it's always caught by the same block check before either of those paths would run. That part stays untested; flagging it rather than papering over it.

```suggestion
    }

    /// Runs a local server that redirects `/` to `location`, so a real
    /// `reqwest::redirect::Attempt` reaches [`ssrf_guarded_redirect_policy`]
    /// -- the type has no public constructor, so this is the only way to
    /// drive it without faking reqwest's internals.
    async fn spawn_redirect_server(location: String) -> String {
        let app = axum::Router::new().route(
            "/",
            axum::routing::get(move || {
                let location = location.clone();
                async move { axum::response::Redirect::temporary(&location) }
            }),
        );
        let listener = tokio::net::TcpListener::bind("127.0.0.1:0").await.unwrap();
        let addr = listener.local_addr().unwrap();
        tokio::spawn(async move {
            axum::serve(listener, app).await.unwrap();
        });
        format!("http://{addr}/")
    }

    #[tokio::test]
    async fn test_redirect_policy_blocks_a_redirect_to_a_disallowed_target() {
        let server_url =
            spawn_redirect_server("http://169.254.169.254/latest/meta-data/".into()).await;
        let client = reqwest::Client::builder()
            .redirect(ssrf_guarded_redirect_policy())
            .build()
            .unwrap();
        let err = client
            .get(&server_url)
            .send()
            .await
            .expect_err("a redirect to a blocked target must fail");
        assert!(err.is_redirect(), "expected a redirect-policy error: {err}");
    }

    #[tokio::test]
    async fn test_redirect_policy_rejects_a_self_redirect_loop_immediately() {
        // A server that redirects to itself is a loopback target, so the
        // per-hop block check (not the separate 10-hop cap, which nothing
        // here can safely reach without a real allowed target to bounce
        // through) is what stops it -- on the very first hop.
        let listener = tokio::net::TcpListener::bind("127.0.0.1:0").await.unwrap();
        let addr = listener.local_addr().unwrap();
        let self_url = format!("http://{addr}/");
        let app = axum::Router::new().route(
            "/",
            axum::routing::get(move || {
                let self_url = self_url.clone();
                async move { axum::response::Redirect::temporary(&self_url) }
            }),
        );
        tokio::spawn(async move {
            axum::serve(listener, app).await.unwrap();
        });

        let client = reqwest::Client::builder()
            .redirect(ssrf_guarded_redirect_policy())
            .build()
            .unwrap();
        let err = client
            .get(format!("http://{addr}/"))
            .send()
            .await
            .expect_err("a loopback self-redirect must be refused, not followed");
        assert!(err.is_redirect(), "expected a redirect-policy error: {err}");
    }
```

muscariello (changes requested):
Consider these changes to keep coverage as it should.

muscariello (approved):
LGTM

Issue #275, opened 2026-09-18
Rust SDK reference server is not ACTS-conformant


muscariello:
The nightly ACTS conformance run flags the Rust SDK as **not conformant**: https://a2aproject.github.io/a2a-itk/dashboard/#/acts/rust

A single failed MUST-level assertion on any transport is enough to fail the whole run, and this run has 9 distinct MUST-level test failures across JSON-RPC/gRPC/HTTP+JSON, each traced to a real bug in `a2a-server`/`a2a`/`a2a-pb` (not a test-harness artifact). Tracking issue for the MUST-level fixes needed to flip conformance to true; all seven need to land, since there's no partial credit.

- [ ] #268 — send_message accepts a follow-up message into an already-terminal task (CORE-SEND-002, CORE-ERR-004)
- [ ] #269 — subscribing to a terminal task reports TaskNotFoundError instead of UnsupportedOperationError (STREAM-SUB-003)
- [ ] #270 — no content-type validation on message parts (CORE-SEND-004)
- [ ] #271 — A2A-Version header is never read or enforced (VER-NEG-001, VER-NEG-002)
- [ ] #272 — malformed JSON-RPC bodies return axum's own 422/400 instead of a 200 + error envelope (CORE-ERR-006, JSONRPC-ERR-002)
- [ ] #273 — ListTasks omits nextPageToken on JSON-RPC/REST (CORE-LIST-003)
- [ ] #274 — task status can be persisted without a timestamp (DM-SERIAL-001)

Once these land, a second, lower-priority pass covers the SHOULD/MAY-level quality failures from the same run (not yet filed as individual issues):

- `AgentCard`'s `description`/`defaultInputModes`/`defaultOutputModes` lack `#[serde(default)]`, so any spec-valid card omitting them fails to parse entirely (CLIENT-CAP-001, CLIENT-AUTH-001)
- REST responses use `application/json` instead of the spec's `application/a2a+json` (`application/problem+json` for errors) (REST-CT-001)
- Agent card endpoint sets no Cache-Control/ETag headers (CARD-CACHE-001)
- Streaming/push-config handlers may have the same capability-vs-existence check-ordering issue as #269 (CORE-CAP-002, PUSH-ERR-001) — needs its own audit
- Task history truncation/suppression behavior doesn't match expectations (CORE-HIST-003, CORE-HIST-005, CORE-HIST-006) — not yet root-caused

PR #284, opened 2026-09-19
fix(a2a-server): backfill missing status timestamp on status update (#274)

### LINKED ISSUE #274: a2a-server: task status can be persisted and returned without a timestamp
ACTS conformance test DM-SERIAL-001: `task.status.timestamp` must always be present. It's missing after `send_message` in this run.

The server does stamp an initial status with `Some(Utc::now())` (`a2a-server/src/handler.rs:659`), but `apply_event_to_task` (`handler.rs:238`) overwrites the whole `TaskStatus` with whatever the `AgentExecutor`'s event carries, with no backfill if that event's status has no timestamp. Any `AgentExecutor` implementation that forgets to set one (including our own ITK conformance fixture, `itk/src/acts.rs:127`) silently produces a spec-violating task.

Fix: in `apply_event_to_task`, backfill `status.timestamp` with `Utc::now()` whenever the incoming event's status has none, so the guarantee holds regardless of what a given `AgentExecutor` supplies.

chopmob-cloud:
## Summary

Two ACTS conformance fixes from the #275 tracker.

**#274 (DM-SERIAL-001): backfill a missing task status timestamp.** `apply_event_to_task` persisted executor events through `save_task` without ensuring the task status carried a timestamp, so a status update (or task snapshot) whose status omitted a timestamp produced a spec-violating task. The backfill now lives in `save_task`, the single persistence chokepoint, so the `StatusUpdate`, `Task`, and `ArtifactUpdate` paths are all covered in one place. A timestamp supplied by the executor is preserved verbatim; only a missing one is filled with `Utc::now()`.

**#275 (SHOULD): default optional AgentCard fields on deserialization.** `AgentCard.description`, `defaultInputModes`, and `defaultOutputModes` lacked `#[serde(default)]`, so the SDK rejected canonical cards that omit them (the ACTS corpus cards do), failing CLIENT-CAP-001 / CLIENT-AUTH-001 where other SDKs pass. Added `#[serde(default)]` to the three fields; a pinned client-parse test that codified the over-strict rejection is updated to assert the card now parses.

## Tests

Added handler tests for the timestamp backfill and preservation on both the status-update and task-snapshot paths, and an AgentCard deserialization test for the omitted fields. `cargo test --workspace --locked`, `cargo fmt --all --check`, and `cargo clippy --workspace --all-targets --locked -- -D warnings` all pass.

Fixes #274. Refs #275.


a2a-server/src/handler.rs
```diff
@@ -213,7 +213,14 @@ async fn send_push_notifications(
...
+    // DM-SERIAL-001: a persisted/returned task status must carry a timestamp.
+    // This is the single persistence chokepoint for executor events, so
+    // backfilling here covers the task-snapshot and status-update paths alike,
+    // while preserving any timestamp the executor already supplied.
+    if task.status.timestamp.is_none() {
+        task.status.timestamp = Some(chrono::Utc::now());
+    }
     match task_store.update(task.clone()).await {
         Ok(_) => Ok(task),
         Err(error) if error.code == error_code::TASK_NOT_FOUND => {
```
muscariello, line 226: The gap is that this backfill only covers what gets *persisted* -- a live subscriber on `send_streaming_message`/`subscribe_to_task` receives the broadcast event directly (see the two call sites in the review summary above), never a reload from the store. Suggesting a helper here that both call sites can use to backfill the event itself before it's persisted *and* published.

```suggestion
}

/// DM-SERIAL-001: backfills a missing status timestamp on the event itself,
/// before it is both persisted (via `apply_event_to_task`/`save_task`) and
/// broadcast to any live subscriber -- `save_task`'s own backfill only
/// covers what gets persisted; a subscriber watching `send_streaming_message`
/// or `subscribe_to_task` receives the broadcast event directly, never a
/// reload from the store, so backfilling only at persistence time still
/// leaves a live subscriber seeing a timestamp-less status.
fn backfill_status_timestamp(event: &mut StreamResponse) {
    let status = match event {
        StreamResponse::Task(task) => &mut task.status,
        StreamResponse::StatusUpdate(update) => &mut update.status,
        StreamResponse::ArtifactUpdate(_) | StreamResponse::Message(_) => return,
    };
    if status.timestamp.is_none() {
        status.timestamp = Some(chrono::Utc::now());
    }
}
```

a2a-server/src/handler.rs
```diff
@@ -3280,6 +3287,172 @@ mod tests {
...
+            }),
+        )
+        .await
+        .unwrap()
+        .expect("task event yields a task");
+
+        assert_eq!(returned.status.timestamp, Some(supplied));
+        let persisted = store.get("t-snapshot-kept").await.unwrap().unwrap();
+        assert_eq!(persisted.status.timestamp, Some(supplied));
+    }
```
muscariello:
```suggestion
    }

    /// DM-SERIAL-001: `save_task`'s backfill only covers what gets persisted
    /// -- a live subscriber receives the broadcast event directly, never a
    /// reload from the store, so the timestamp must also be backfilled on
    /// the event itself before it is published.
    #[tokio::test]
    async fn test_streamed_task_event_has_a_backfilled_timestamp() {
        use futures::StreamExt;
        let handler = make_handler();
        let params = ServiceParams::new();
        let req = SendMessageRequest {
            message: make_message(),
            configuration: None,
            metadata: None,
            tenant: None,
        };
        let mut stream = handler.send_streaming_message(&params, req).await.unwrap();
        let event = stream.next().await.unwrap().unwrap();
        match event {
            StreamResponse::Task(task) => assert!(
                task.status.timestamp.is_some(),
                "a live-streamed task event must carry a backfilled timestamp"
            ),
            other => panic!("unexpected event: {other:?}"),
        }
    }
```

muscariello (commented):
Verified both fixes: the `save_task` backfill is correctly placed at the single persistence chokepoint (all three `apply_event_to_task` branches route through it), and the `AgentCard` `#[serde(default)]` fix is spec-correct -- checked against `A2A/specification/json/a2a.json` directly: `description` has an explicit `"default": ""` and none of the three fields appear in any `required` list. Ran the full workspace suite, `fmt`, and `clippy -D warnings` locally; all clean, matching what CI already shows.

One real gap, confirmed by actually running it, not just reading the code: `save_task`'s backfill only covers what gets *persisted* and what `send_message` returns after reloading from the store. A client on `send_streaming_message`/`subscribe_to_task` gets the *broadcast* event directly (`ActiveExecution::publish` sends the original event; the corrected task only updates internal snapshot state). I reproduced it with the existing `EchoExecutor` fixture -- before any further change, a live-streamed event's `status.timestamp` came back `None`. The exact same pattern also exists in `cancel_task`'s own live-publish path -- same broadcast-vs-reload gap, same fix applies.

The suggestion below (on `save_task`) adds a helper that backfills the event itself, before it's both persisted and published, plus a test pinning the streaming case. Its two call sites are outside this PR's diff, so GitHub won't let me anchor a suggestion there -- they're below as plain snippets instead. Verified: 204 tests pass (203 existing + 1 new), `fmt`/`clippy -D warnings` clean, no regressions. The `cancel_task` call site is fixed by the same mechanism but I didn't build a dedicated live-subscription-during-cancel test for it -- flagging that so it's not overclaimed as independently verified the way the streaming case is.

**Call site 1** -- `drive_execution` (the `send_streaming_message`/`subscribe_to_task` path), where `match result { Ok(event) => {` currently reads:

```rust
        match result {
            Ok(mut event) => {
                backfill_status_timestamp(&mut event);
                match apply_event_to_task(runtime.task_store.as_ref(), current_task.clone(), &event)
                    .await
```

**Call site 2** -- `cancel_task`'s own event loop, where `match event { Ok(event) => {` currently reads:

```rust
        while let Some(event) = stream.next().await {
            match event {
                Ok(mut event) => {
                    backfill_status_timestamp(&mut event);
                    let updated_task =
                        apply_event_to_task(self.task_store.as_ref(), current_task.clone(), &event)
                            .await?;
```

muscariello (changes requested):
Please check the few comments. Overall it's ok. Thanks for your contribution !

muscariello (approved):
LGTM

